PR #2, opened 2015-12-18
001: initial C++ Streams proposal

indutny:
cc @trevnorris

### DIFF
--- a/001-public-stream-base.md
+++ b/001-public-stream-base.md
@@ -1,0 +1,166 @@
+# Make C++ Stream APIs public
+
+Author: Fedor Indutny <fedor@indutny.com>
+Status: draft
+Date: Thu Dec 17 2015 19:39:46 GMT-0500 (EST)
+
+## Description
+
+Currently we use so called `StreamBase` APIs through the core to implement
+various performance-critical networking protocols with maximum efficiency.
+It is evident that this performance improvement could be beneficial for
+user-land C++ addons too.
+
+I propose to make this APIs public, introducing C++ interface similar to the
+following one (simplified):
+
+## Interface
+
+```C++
+class Stream {
+ public:
+  //
+  // Low-level part
+  //
+
+  // Perform protocol-specific shutdown (EOF)
+  virtual int DoShutdown(ShutdownWrap* req_wrap) = 0;
+
+  // Write data asynchronously
+  virtual int DoWrite(WriteWrap* w,
+                      uv_buf_t* bufs,
+                      size_t count,
+                      uv_stream_t* send_handle) = 0;
+
+  // **(optional)**
+  // Try to write provided data immediately without blocking, if not
+  // possible to do - should return `0` if no data was written, should decrement
+  // `count` and change pointers/length in `bufs` if some data was written.
+  virtual int DoTryWrite(uv_buf_t** bufs, size_t* count);
+
+  //
+  // High-level part
+  //
+
+  // Cast to the deepest class. Just a stub, probably does not need to be
+  // here.
+  virtual void* Cast() = 0;
+
+  // Return `true` or `false` depending on liveness of the underlying
+  // resource. If Stream is not alive - all operations on it will result in
+  // `EINVAL`
+  virtual bool IsAlive() = 0;
+
+  // Return `true` if stream is currently closing. Closed streams should return
+  // `false` when calling `IsAlive`
+  virtual bool IsClosing() = 0;
+
+  // **optional**, if `true` - `send_handle` may be not `nullptr` in `DoWrite`
+  virtual bool IsIPCPipe();
+
+  // **optional**, return fd
+  virtual int GetFD();
+
+  // Start reading and emitting data
+  virtual int ReadStart() = 0;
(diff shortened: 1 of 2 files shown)


001-public-stream-base.md
```diff
@@ -0,0 +1,166 @@
+# Make C++ Stream APIs public
+
+Author: Fedor Indutny <fedor@indutny.com>
+Status: draft
+Date: Thu Dec 17 2015 19:39:46 GMT-0500 (EST)
```
trevnorris: Consider doing this how is done in LEPS (https://raw.githubusercontent.com/libuv/leps/master/001-remove-unix-support.md). As this is now all these entries will be placed on the same line.

indutny: Like this?

001-public-stream-base.md
```diff
@@ -0,0 +1,166 @@
...
+  // Write data asynchronously
+  virtual int DoWrite(WriteWrap* w,
+                      uv_buf_t* bufs,
+                      size_t count,
+                      uv_stream_t* send_handle) = 0;
+
+  // **(optional)**
+  // Try to write provided data immediately without blocking, if not
+  // possible to do - should return `0` if no data was written, should decrement
+  // `count` and change pointers/length in `bufs` if some data was written.
```
trevnorris, line 38: Can we include a helper function or whatnot for recalculating the `uv_buf_t**`? We have the logic already in `StreamWrap::DoTryWrite()`, and it'd be a shame for everyone to need to implement basically the exact same on their own.

001-public-stream-base.md
```diff
@@ -0,0 +1,166 @@
...
+  //
+
+  // Perform protocol-specific shutdown (EOF)
+  virtual int DoShutdown(ShutdownWrap* req_wrap) = 0;
+
+  // Write data asynchronously
+  virtual int DoWrite(WriteWrap* w,
+                      uv_buf_t* bufs,
+                      size_t count,
+                      uv_stream_t* send_handle) = 0;
```
trevnorris, line 33: Will all these APIs be required to return correct `UV_E*` error codes?

001-public-stream-base.md
```diff
@@ -0,0 +1,166 @@
...
+  virtual void* Cast() = 0;
+
+  // Return `true` or `false` depending on liveness of the underlying
+  // resource. If Stream is not alive - all operations on it will result in
+  // `EINVAL`
+  virtual bool IsAlive() = 0;
+
+  // Return `true` if stream is currently closing. Closed streams should return
+  // `false` when calling `IsAlive`
+  virtual bool IsClosing() = 0;
```
trevnorris, line 56: Just curious. Are streams that are in the process of closing alive?

indutny, line 56: nope, they are not. Perhaps, wording could be better here, but this is what I was trying to say in the second sentence.

trevnorris, line 56: Guess what I'm wondering is when is it necessary to know specifically if it's closing. Instead of only whether it's alive or not?

001-public-stream-base.md
```diff
@@ -0,0 +1,166 @@
...
+
+  // **optional**, if `true` - `send_handle` may be not `nullptr` in `DoWrite`
+  virtual bool IsIPCPipe();
+
+  // **optional**, return fd
+  virtual int GetFD();
```
saghul, line 62: We have some `uv_os_fd_t` thingy, which is int on Unix and HANDLE on Windows, FWIW.

001-public-stream-base.md
chrisdickinson, line 56: In event of an error, knowing whether or not to attempt to close the stream (in case the error was triggered because of the close request)?

chrisdickinson:
This might not be in-scope for this NEP, but one of the more interesting attributes of WHATWG streams is the ability to do brand-checking during `readable.pipe(writable)` and skip a round trip into JS if both the readable and writable sides can communicate at the C++ layer. It'd be interesting to expose this functionality through the public stream implementation — so that if a non-handle-backed Stream were piped to a handle-backed stream, things would still work, but if two handle-backed streams were `.pipe()`'d together, they could communicate directly.

indutny:
Sorry for delay, I took it to consider things more carefully. Turns out, I agree with @bnoordhuis , adding this won't really create more value than problems. Let's close this for now.

Thank you for your time, and sorry for wasting it on this. :disappointed:

trevnorris:
@indutny :( The concept was good. Hopefully we can figure out how to get something like this into core in the future.

indutny:
We have this in core already :) I'm not suggesting to remove it, just don't think that making it public would be beneficial.

trevnorris:
Sorry. should have read "as public API in core"

trevnorris:
Forgot that even rejected proposals are still merged. Landed in 9faa231.

mcollina:
What's the status of this proposal in core? Was it exposed?

indutny:
I believe the proposal was rejected. The APIs are not exposed.

PR #4, opened 2016-01-16
Buffer.from/Buffer.alloc/Buffer.zalloc/Buffer() soft-deprecate

jasnell:
Refs: https://github.com/nodejs/node/pull/4682
Refs: https://github.com/nodejs/node/issues/4660

### DIFF
--- a/000-index.md
+++ b/000-index.md
@@ -1,1 +1,3 @@
 # Node.js Enhancement Proposals
+
+* [Public C++ Streams][001-public-stream-base.md]
--- a/001-public-stream-base.md
+++ b/001-public-stream-base.md
@@ -1,0 +1,166 @@
+| Title  | Make C++ Stream APIs public |
+|--------|-----------------------------|
+| Author | @indutny                    |
+| Status | REJECTED                    |
+| Date   | 2015-12-17 19:39:46         |
+
+## Description
+
+Currently we use so called `StreamBase` APIs through the core to implement
+various performance-critical networking protocols with maximum efficiency.
+It is evident that this performance improvement could be beneficial for
+user-land C++ addons too.
+
+I propose to make this APIs public, introducing C++ interface similar to the
+following one (simplified):
+
+## Interface
+
+```C++
+class Stream {
+ public:
+  //
+  // Low-level part
+  //
+
+  // Perform protocol-specific shutdown (EOF)
+  virtual int DoShutdown(ShutdownWrap* req_wrap) = 0;
+
+  // Write data asynchronously
+  virtual int DoWrite(WriteWrap* w,
+                      uv_buf_t* bufs,
+                      size_t count,
+                      uv_stream_t* send_handle) = 0;
+
+  // **(optional)**
+  // Try to write provided data immediately without blocking, if not
+  // possible to do - should return `0` if no data was written, should decrement
+  // `count` and change pointers/length in `bufs` if some data was written.
+  virtual int DoTryWrite(uv_buf_t** bufs, size_t* count);
+
+  //
+  // High-level part
+  //
+
+  // Cast to the deepest class. Just a stub, probably does not need to be
+  // here.
+  virtual void* Cast() = 0;
+
+  // Return `true` or `false` depending on liveness of the underlying
+  // resource. If Stream is not alive - all operations on it will result in
+  // `EINVAL`
+  virtual bool IsAlive() = 0;
+
+  // Return `true` if stream is currently closing. Closed streams should return
+  // `false` when calling `IsAlive`
+  virtual bool IsClosing() = 0;
+
+  // **optional**, if `true` - `send_handle` may be not `nullptr` in `DoWrite`
+  virtual bool IsIPCPipe();
+
+  // **optional**, return fd
+  virtual int GetFD();
+
+  // Start reading and emitting data
+  virtual int ReadStart() = 0;
+
+  // Stop reading and emitting data (immediately)
+  virtual int ReadStop() = 0;
+
+ protected:
+  // One of these must be implemented
+  virtual AsyncWrap* GetAsyncWrap();
+  virtual v8::Local<v8::Object> GetObject();
+};
+```
+
+## Public APIs
+
+Public facing APIs have two sides: JS and C++.
+
+### JS
+
+```js
+stream.fd; // stream's fd or -1
+stream._externalStream;  // External pointer to the deepest C++ class
+stream.readStart();
+stream.readStop();
+stream.shutdown(req);
+stream.writev(req, [ ..., chunk, size, ... ]);
+stream.writeBuffer(req, buffer);
+stream.writeAsciiString(req, string);
+stream.writeUtf8String(req, string);
+stream.writeUcs2String(req, string);
+stream.writeBinaryString(req, string);
+```
+
+### C++
+
+All of the C++ interface methods may be called. Additionally following methods
+are available:
+
+```C++
+class Stream {
+ public:
+  //
+  // Low-level APIs
+  //
+
+  // Invoke `after_write_cb`
+  inline void OnAfterWrite(WriteWrap* w);
+
+  // Invoke `alloc_cb`
+  inline void OnAlloc(size_t size, uv_buf_t* buf);
+
+  // Invoke `read_cb`
+  inline void OnRead(size_t nread,
+                     const uv_buf_t* buf,
+                     uv_handle_type pending = UV_UNKNOWN_HANDLE);
+
+  // Override and get callbacks
+  inline void set_after_write_cb(Callback<AfterWriteCb> c);
+  inline void set_alloc_cb(Callback<AllocCb> c);
+  inline void set_read_cb(Callback<ReadCb> c);
+  inline Callback<AfterWriteCb> after_write_cb();
+  inline Callback<AllocCb> alloc_cb();
+  inline Callback<ReadCb> read_cb();
+
+  //
+  // High-level APIs
+  //
+
+  // Add JS API methods to the target `FunctionTemplate`
+  // NOTE: `flags` control some optional methods like: `shutdown` and `writev`
+  static inline void AddMethods(Environment* env,
+                                v8::Local<v8::FunctionTemplate> target,
+                                int flags = kFlagNone);
+
+  // Emit data to the JavaScript listeners
+  void EmitData(ssize_t nread,
+                v8::Local<v8::Object> buf,
+                v8::Local<v8::Object> handle);
+
+  // Mark stream as consumed
+  inline void Consume();
+
+  // TODO(indutny): add this to node.js core
+  inline void Unconsume();
+};
+```
+
+## Putting things together
+
+This APIs could be used for implementing high-performance protocols in a
+following way:
+
+Each TCP socket has a `_handle` property, which is actually a C++ `Stream`
+instance. One may wrap this `_handle` into a custom C++ `Stream` instance, and
+replace the `_handle` property with it. This stream could override `read`,
+`alloc`, `after_write` callbacks to hook up the incoming data from the wrapped
+stream.
+
+Calling `DoWrite` and friends on the wrapped stream will result in performing
+actual TCP writes.
+
+It is easy to see that it could be wrapped this way several times, creating
+levels of abstraction without leaving C++!


indutny:
You have my +1 on this

ChALkeR:
+1 for the primary goal, except for the new method names. **That is not a nitpick, I consider those to be very important.**
_I mentioned that a few times already, but making sure that this discussion does not miss it._

With the currently included method names, this proposal fixes only problem 2, but does not fix problem 1 from [my note](https://github.com/ChALkeR/notes/blob/master/Lets-fix-Buffer-API.md#is-there-a-problem) — inexperienced users will still use an unsafe method in a dangereous way without reading the docs first, and there will be more bugs in the community modules. See also [q/a](https://github.com/ChALkeR/notes/blob/master/Lets-fix-Buffer-API.md#qa) question 18.

Also, I am a bit concerned about the secondary goal — even if that change is _sometime_ pushed to v0.12.n+1, users that would use v0.12.n will be endangered. That does not change over time, replace «n» with the latest released version to that time (and «v0.12» with any branch). _This point is valid until `Buffer(number)` is hard-deprecated._ Also, the `--unitialized-buffers` command line option would work as a kill switch if any author of any lib assumes that `Buffer(number)` will be zero-filled, and they would assume that, once this change is brought to the public. I can already imagine «Node.js zero-fills `Buffer(number)` now» news and tweets.

rwaldron:
Is the intention to define `Buffer.from` such that it overrides the built-in class-side `from` method that `Buffer` [inherits](https://github.com/nodejs/node/blob/66b9c0d8bdb7f9fb542f53c2c18bacd5bbca272d/lib/buffer.js#L47-L68) from `Uint8Array`? Take a look: 

``` js
function Buffer() {} 

console.log(Buffer.from === undefined); // true

// See buffer.js#L47-L71, nodejs/node @ 66b9c0d (linked above) 
Object.setPrototypeOf(Buffer.prototype, Uint8Array.prototype);
Object.setPrototypeOf(Buffer, Uint8Array);

console.log(Buffer.from === undefined); // false
console.log(Buffer.from === Uint8Array.from); // true
```

See also: 
- [%TypedArray%.from](https://tc39.github.io/ecma262/#sec-%typedarray%.from)
- [TypedArrayFrom](https://tc39.github.io/ecma262/#sec-typedarrayfrom)

jasnell:
Yes, it overrides the built in from that is inherited from Uint8Array.

rwaldron:
> Yes, it overrides the built in from that is inherited from Uint8Array.

Thank you for clarifying these semantics. If this proposal is accepted, the docs should make mention that `Buffer.from` is an override and does not offer the mapping function argument.

glenjamin:
I would also like to add that I think it should be made marginally "harder" to use `alloc` than `zalloc`.

That is: If a developer only glances at the API without properly reading docs, it would be beneficial if they picked the method which is safest. Those who need / want the extra performance can find it by reading the documentation more closely. Someone who wants to allocate some memory may not read every buffer method - they'll likely stop when they find one which fits the bil

Something like `zalloc -> alloc` + `alloc -> allocRaw` would acheive this, or having a scary warning very clearly in the docs for the `alloc` which doesn't zero.

mafintosh:
Looks good! This would have solved the security problems we've found in bittorrent-dht and ws. Thanks for turning this into an EPS @jasnell.

joepie91:
> @glenjamin: I would also like to add that I think it should be made marginally "harder" to use alloc than zalloc.
> 
> That is: If a developer only glances at the API without properly reading docs, it would be beneficial if they picked the method which is safest. Those who need / want the extra performance can find it by reading the documentation more closely. Someone who wants to allocate some memory may not read every buffer method - they'll likely stop when they find one which fits the bil

+1 on this. Again, I feel that 'unsafe' is the most accurate and beginner-friendly way to convey this, rather than 'raw' or other more 'neutral' terms.

wbl:
This proposal does not actually fix the bug. new Buffer(variable) will continue to result in information disclosure vulnerabilities until existing code is changed. I still don't understand why this behavior change isn't being made.

alliejp-tv:
> @joepie91: I feel that 'unsafe' is the most accurate and beginner-friendly way to convey this, rather than 'raw' or other more 'neutral' terms.

**+1**. `buffer.alloc()` and `buffer.unsafealloc()` would most definitely keep people away from the second option unless they knew what they were doing.

feross:
Nice work. This solves the core problem that affected ws and bittorrent-dht, which is `Buffer(variable)` getting tricked into taking a number argument.

rvagg:
@nodejs/ctc this is probably a good place to move technical discussion for now.

@jasnell can we have clarification on what "soft-deprecation" means? Is it doc-deprecation? With "hard-deprecation" meaning `util.deprecate()`?

_(apologies for the long prose to follow but this seems like the appropriate to finally lay out my current thoughts)_

Backporting to v0.10, v0.12, and v4 could actually make this more painful than it should be, we'll then have a pretty big feature-difference between patch-versions. I'm guessing the intention is to make the new factory functions available to everyone using Node.js but that makes the assumption that the majority of users are using the latest release in each line, which is not the case at all. I've seen companies justify the ongoing use of older v0.10 in production with typical risk-aversion combined with claims that being behind load-balancers and TLS terminators protect them from most of the security flaws that would normally force people to upgrade. There's also situations where Node is being used completely disconnected and therefore less at risk to security vulnerabilities. Even beyond commercial use, `node` is used as part of a utility toolchain so widely now and many `node` users either don't care or don't know what version they are using (if they know they are using it at all). I don't endorse any reasons people give for not upgrading when we have patch-releases but it happens all the time and will continue to happen. If we end up with the npm ecosystem shifting to the new factory methods when they are only available in the latest of a release line we're going to have a significant amount of pain. We've had a couple of historical instances like this that we should learn from. The last, most painful one that I recall is the introduction of `^` in npm 1/2 way through v0.10, leaving older v0.10 and v0.8 installs without it. When it was turned on as the default for `npm install -S` we had _soo_ much breakage. The number of bug reports that many of us has to deal with in our various projects related to users' npm being unable to parse the `^` in `"dependencies"` was phenomenal. And consider here that the solution for many people was to just `npm install npm -g` to upgrade (except when npm itself broke old npm for the same reason!) or simply installing the latest v0.10. The demonstration here was clear that we can't use "it's in the latest release" as a mechanism to solve anything.

I'd _really_ prefer we stick to semver as much as possible as it solves a lot of these kinds of problems. At this stage I think my preference, if we are determined to act, would be something like:
1. introduce new factory methods on v5 and and a v4 semver-minor
   - aside: I don't particularly like `zalloc` and `alloc`, I also don't like the word "safe" in these, but that seems like a minor discussion when we move past the big issues
2. document deprecation of the constructor in all cases, make it super-clear in the `Buffer` docs what the issues are
3. _(maybe!)_ introduce zero-fill-by-default for the constructor and and cmdline flag to turn it off in all active release lines, including v0.10 and v0.12
4. at the end of 2016, do a `util.deprecate()` on the constructor, v0.10 and v0.12 will be EOL by then
5. _(much later)_ remove the constructor

I don't think the accidental-`number` problem is something we can hit straight away except through education. There's too much breakage in taking forceful action while v0.10 and v0.12 are still active. It may be safe enough to backport a warning on `new Buffer(number, enc)` to both v0.10 and v0.12, although this doesn't exactly solve the problem.

To re-state what I said above, introducing factory methods to v0.10 and v0.12 will not solve the version discrepancies because it assumes people use the latest, by not adding the factory methods to those versions people are much more likely to write and use feature-detection mechanisms, which is good, and we could even encourage this in the docs (`var buf = Buffer.zalloc ? Buffer.zalloc(1024) : new Buffer(1024).fill(0)`). This all goes away at the end of 2016, like saying goodbye to IE6.

While I accept that we have a usability problem here and that core has to take some responsibility for the way people use Node, I don't accept that we should treat such problems in the same way that we treat core security vulnerabilities. Hence my insistence on taking a longer view on this with v0.10 and v0.12 to prioritise greater ecosystem stability over the possibility of ecosystem packages containing similar API misuse to what we are seeing. Vulnerabilities in individual npm packages are almost always going to be of much lesser impact that vulnerabilities in core and they are also much easier to patch and distribute without the kind of breakage and uncertainty that happens when we do it with core.

And lastly, something I keep on repeating, it's essential that we do our best to educate people to think about Node as a **server-side programming platform** where evil can wreak havoc if you don't think through what you're doing carefully and be attune to the kinds of common vulnerabilities that can be created. Node.js is a platform that attracts users who are used to a browser sandbox and are not accustomed to thinking as carefully about input sanitisation and careful interaction with system resources. This is going to be a continual theme (TrendMicro being the current big news on this front) and while we can tackle it in part by making the API safer we also need to accept that nannying users risks lulling them into a perception that Node.js is more of a sandbox than it is. We _must_ accept a caveat emptor approach.

ChALkeR:
> @jasnell can we have clarification on what "soft-deprecation" means? Is it doc-deprecation? With "hard-deprecation" meaning util.deprecate()?

This is what I meant in all my messages, and I hope that this is what it means here.

> Backporting to v0.10, v0.12, and v4 could actually make this more painful than it should be … I'd really prefer we stick to semver as much as possible

v0.10 and v0.12 are not semver, but I understand your concerns. Though, note that this EPS does not describe backporting this feature to v0.10/v0.12, and I personally view that as a separate question. If you want to know my opinion on this — backporting the new API methods to v0.10 and v0.12 would assert faster ecosystem migration to the safer methods, _and_ will motivate people who are using older patch releases of v0.10 and v0.12 to update to the latest patch version (which I also view as a positive change), but _only when they will update their libraries_. Also note the the popular libraries could have these concerns themselves, so they could intoduce migration to the new Buffer api in a semver-major version of the library. Also, merging this to all the supported branches will allow  _new_ libraries to use the safer API from the start. Still, I view this discussion as controversial and less important, thus I do not want to discuss this in the current pull, unless backporting new API changes is included to the EPS itself.

> I don't think the accidental-number problem is something we can hit straight away except through education.

Of course it wouldn't be hit straight away, partially because the old methods would be only soft-deprecated for quite a while, partially because people generally don't react very fast and there are a lot of packages that need to be changed. The primary goal of this is to provide safer API for use in the future (as it seems to me), 

> aside: I don't particularly like `zalloc` and `alloc`, I also don't like the word "safe" in these, but that seems like a minor discussion when we move past the big issues

Again: I view this to be **very** important, because choosing wrong method names would fix only the second problem (accidental `Buffer(number)` hitting instead of `Buffer(value)`), but not the first one (inaccurate uninitialized `Buffer(number)` usage, even for not performance-critical methods), but will leave us in a situation where that is even less likely to be fixed than it is now — because _renaming_ methods would break more things than introducing new ones.

> _(maybe!)_ introduce zero-fill-by-default for the constructor and and cmdline flag to turn it off in all active release lines, including v0.10 and v0.12

Please, no. While new API will cause the previous versions (in the branches it gets backported to) to crash (if libraries start relying on new API), this change will make them leak uninitialized buffers unnoticed (if the libraries start relying on zero-fill-by-default), which is far far worse. Also note that you can't stop them from relying on zero-fill-by-default once it's introduced — almost noone will want to have _two_ zero-fills.

> And lastly, something I keep on repeating, it's essential that we do our best to educate people to think about Node as a _server-side programming platform_ where evil can wreak havoc if you don't think through what you're doing carefully and be attune to the kinds of common vulnerabilities that can be created.

Strongly agreed, moreover, there are several more specific education-related issues that I wanted to discuss after this gets fixed (and one other change).

> and while we can tackle it in part by making the API safer we also need to accept that nannying users risks lulling them into a perception that Node.js is more of a sandbox than it is

The problem is (as you already said) that they already have this perception, and there are no reasons not to guard them from _common pitfalls_ (like the Buffer API), leaving more dangereous methods to be the less obvious to use ones and reachable by more experienced developers. Btw, if your statement was in protection of not changing the API, then it was a very bad reasoning for not making the ecosystem safer — those effects are incomparable, `Buffer()` API is hurting everyone _right now_. If I misread — please, excuse me.

joepie91:
@rvagg Well, @ChALkeR has said most of it, but just to weigh in on the naming issue once again: it is _critical_ that the `unsafe` terminology or a similarly dangerous-looking word is used, to steer people away from the dangerous option unless they know what they are doing.

That being said, I do not particularly care about using `safe` anywhere - so having a `Buffer.alloc` and `buffer.unsafeAlloc` method, for example, would be completely okay as far as I'm concerned. I do understand that it may be a bad idea to call something `safe` when there may be other possible issues, so this might be the best way to approach the naming.

glenjamin:
> To re-state what I said above, introducing factory methods to v0.10 and v0.12 will not solve the version discrepancies because it assumes people use the latest, by not adding the factory methods to those versions people are much more likely to write and use feature-detection mechanisms, which is good, and we could even encourage this in the docs (var buf = Buffer.zalloc ? Buffer.zalloc(1024) : new Buffer(1024).fill(0)). This all goes away at the end of 2016, like saying goodbye to IE6.

Do you think this is a good case for doing something similar to `readable-stream` from the 0.8 to 0.10 transition? Is that something which was considered successful and/or a good idea with hindsight?

ie. a widely used userland module which provides these semantics, using native methods where available.

joepie91:
> Do you think this is a good case for doing something similar to readable-stream from the 0.8 to 0.10 transition? Is that something which was considered successful and/or a good idea with hindsight?

I realize you weren't asking me specifically, but: it's actually not entirely clear to me what the purpose of the `readable-stream` module is, or rather, whether it still plays a relevant role in the current state of Node.js. I have seen many others voice the same.

Aside from that, many people still perceive a 'cost' to using a userland module, even if there isn't any such cost. "I don't want to depend on third-party modules for this" is still a commonly heard argument in #Node.js, and it is still an extra hurdle for people to take. The same concern applies to "recommending people to write a polyfill".

jasnell:
@rvagg ... thank you for the lengthy review... I think you and I are on the same page.

@ChALkeR @joepie91 and others... the key challenge with the `unsafe` terminology is that the method itself is not unsafe, only the incorrect usage of it is. There are perfectly legitimate and _quite safe_ uses for allocating uninitialized memory. The goal here is to come up with a compromise that is a good balance between all of the various points of view, I'm quite certain that we're not going to make everyone happy in this case but we need to try to settle on names that are good enough to address the core problem. 

Currently I'm thinking: 
- rename `zalloc` to `alloc` to ensure that it (a) comes first in the alphabetical listing 
- rename `alloc` to `ualloc` for "uninitialized allocation" ... this will make it appear to be the non-default choice and will ensure that it comes later in the alphabetical listing.

joepie91:
> the key challenge with the unsafe terminology is that the method itself is not unsafe, only the incorrect usage of it is.

No, it isn't. An unitialized memory allocation is _always_ unsafe from a memory safety point of view. That you can mitigate the issue doesn't change that the operation itself is unsafe. That there are 'safe uses' also doesn't change that. None of that concerns the operation itself.

EDIT: And if you (general you, not personal you) are sufficiently competent to understand how to use an uninitialized allocation, then you are also capable of understanding that "unsafe" concerns memory safety. Yes, this makes it "look scary" - that is precisely the point. There's no technical argument not to.

glenjamin:
@jasnell  I think the rust approach to this is quite nice: https://doc.rust-lang.org/book/unsafe.html

The general theme is that pretty much everything else is safe, and unchecked usage of this function is unsafe. Therefore the method gets marked unsafe, and you are required to wrap it in some safety.

Node.js and rust have quite different target audiences and use-cases, but I think in this case there is some overlap.

Another argument for making them more different is that if you're skimming through code, a single letter difference isn't always immediately visible. Something that can potentially expose runtime memory should look visually distinct, and jump out at the reader.

As with everything else I've written, this is all IMO.

jasnell:
To be clear, _personally_ I have no problem with using `unsafe` in the term. Others have raised objections to that language in the various threads and my goal is to come up with a compromise that works _for the most part_.

joepie91:
@jasnell I understand - I believe you've explicitly remarked on representing objections from others in the previous thread, so I've been more or less assuming that throughout the course of the threads :) Hence my attempt to re-explain why the use of the `unsafe` terminology is appropriate here.

If there are people who still object  to the usage of the term `unsafe`, I'd suggest they raise their objection in this thread personally, and preferably that they also address the distinction I've made regarding memory safety.

I've made this argument several times now, but technical counter-arguments have not been forthcoming - and it seems silly to block the decision on an objection that's not backed by a technical argument, or which may even already have been retracted because of the "memory safety" distinction.

ChALkeR:
@jasnell 

I have no _hard_ feeling on the need of the `unsafe` word in the method name itself, i.e. `alloc/ualloc` IMO is _much_ better than `zalloc/alloc`, but `alloc/unsafeAlloc` or `alloc/uninitializedAlloc` is _slightly_ better than `alloc/ualloc`. 

That is valid if and only if the deprecation message (both the soft-deprecation docs one and the future hard-deprecation code one) of `Buffer(number)` links only to `Buffer.alloc()` and not to `Buffer.ualloc()` (so that everyone not reading the docs carefully will choose the safe method), and `Buffer.alloc` documentation has a reference to `Buffer.ualloc`, marking it as not being zero-filled and potentially unsafe (i.e. «double check that you are sure what you are doing and if you need this» type of warning).

Let's wait for what the opponents of the `unsafe` word in the method name will say.

glenjamin:
I really like `alloc/uninitializedAlloc`, it ticks all my boxes!
- visually distinct from the safe version
- safer one looks nicer
- does what it says on the tin

And for bonus points, doesn't include `unsafe`.

:+1:

jasnell:
Yes but `uninitializedAlloc` has soo many characters to type... lol... I'd be much happier with `unsafeAlloc` :-)

trevnorris:
I take offence at your insistence that usage of uninitialized memory is _always_ unsafe. Take the following:

``` js
var b = Buffer.alloc(n).fill('abc');
```

or maybe this:

``` js
var b = Buffer.alloc(n);
fs.read(fd, b, 0, b.length, 0, cb);
```

or possibly I'm just not being lazy and actually keep track of what I'm doing. The rebuttal of having a moment of potentially "unsafe" memory is pointless. There's a potential for anything to happen if a developer isn't careful. Allowing unchecked variables to be processed from an outside source is a pretty big mess up regardless of what the results are. Don't tell me what's unsafe. Just tell me what's technically happening.

That said, `alloc()` and `allocRaw()` seems appropriate enough. While `allocUninitialized()` is more technically correct I'd smack someone if I had to be typing that in constantly.

trevnorris:
@jasnell There was a mess-up in how these EPS's are supposed to be processed. The document is supposed to start with `XXX` and it's numbered based on what order it lands. I'm going to actually land Fedor's first one today and get this kicked off.

wbl:
What the programer did was assume that new Buffer(object) always serializes a representation of object. That doesn't happen when object is a number. Pointless pontificating about "server-side can unleash evil" and the as-yet unsuccessful idea of sufficiently smart programmers doesn't change the underlying problem, which can be fixed by a simple change that requires no action by any programmer.

The current approach in this branch doesn't fix this. Instead it creates new constructors, which existing code will need to be rewritten to use. In the meantime existing code remains vulnerable. The funny overload behavior will remain.

trevnorris:
@jasnell Also before this is merged the API documentation (not full, just technical overview) should be placed directly in this EPS.

ChALkeR:
@wbl I think everyone here is aware of the downsides of this proposal, but do you have any better ideas? Just make sure that you read all the previous discussion and links from that EPS in advance and try to somehow avoid the downsides of alternate proposals.

If you are trying to advocate the zero-fill-all-the-buffers approach — please read my comments in this tread, my [note](https://github.com/ChALkeR/notes/blob/master/Lets-fix-Buffer-API.md) (especially «is there a problem» part, «things that should be noted» part, and q/a entry 3), then try to answer how would you deal with all the described problems.

joepie91:
> @trevnorris: I take offence at your insistence that usage of uninitialized memory is always unsafe. Take the following:
> 
> or maybe this:

Both of those Buffer allocations are unsafe. Again, I am referring to the operation of allocating a Buffer, not to what you do with it afterwards. In other words, it _tells you what is technically happening_ - it's making an unsafe allocation, and then fills/overwrites it with data.

I also still do not see a technical argument here as to why `unsafe` naming is _undesirable_. I only see an argument why you feel it's _not necessary_, which is something very different.

EDIT: And just for the sake of completeness:

> There's a potential for anything to happen if a developer isn't careful. 

This is an ancient excuse for insecure-by-default code, and it does not help anybody. It just harms people by increasing the likelihood of vulnerabilities, whether due to mistakes or lack of knowledge. "Let's not bother to try and make this more secure, because there will always be something else that's insecure" is still just as invalid an argument as it has always been.

jasnell:
Ok, I've pushed an update to the proposal PR just to reflect the ongoing process of this discussion.

The `zalloc`/`alloc` methods are renamed to `alloc`/`allocRaw` and the documentation is expanded (hopefully) make the distinction between the two clearer.

I also slightly tweaked the signature of `alloc` from `alloc(size)` to `alloc(size[, fill])`, where `fill` must be a string specifying what to fill the buffer with. When `fill` is passed, a non-pooled zero-filled buffer is returned, if `fill` is specified, it is equivalent to doing `Buffer.allocRaw(size).fill(fill)`.

Again, this update simply reflects the ongoing discussion...

wbl:
I'm not opposed to API evolution. But there is no reason the behavior of new Buffer(number) can't be fixed, while we wait for code to be updated to use the new APIs. I think the arguments being made against this are just wrong, especially the ones about security havok in the 5.x series. Just backport already/realise they are already getting screwed.

jasnell:
@trevnorris ... OK, I'll work up the changes to the EPS PR a bit later on this afternoon

ChALkeR:
@wbl

The majority of all developers (including the majority of developers from Node.js ecosystem) do not take security seriously enough. The attitude usually goes from «don't bug me with uninteresting things, I'm trying to do real work here» / «management wants features» to «security is good, I will happily fix vulnerabilites when someone finds them, but I don't like to think hard about it in advance». This is in no way unique to Node.js ecosystem — pretty much everywhere it's that way. Add the fact that security is **hard**, and it often requires a careful and dedicated consideration to think about the possible consequences of decisions being made. I am not saying anything new here, just preparing you for the next conclusion.

The above facts will cause the zero-fill-all-the-buffers change to be immediately intercepted as «`Buffer(number)` is now safe, we don't need to zero-fill it anymore!» by a major part of developers. New package authors will ignore the recommendation to cast `.fill(0)`, they will think «I don't need no old versions». Even those who _did_ have `.fill(0)` will remove it, because in eyes of many of them the speed profit would be more valuable than «some old versions of Node.js considerations» that do not break things.

Add here  the fact that new versions (of everything, but I currently refer to Node.js here) do not instanly reach various setups even if the changes would be introduced to all the branches, but do that quite slowly, and people are very likely to bump versions of libs (or even getting the shiny new ones) without updating Node.js itself (which often brings nothing in their eyes, especially if they are not going to use new features and rewrite their stuff).

Add here the fact that the zero-fill-all-the-buffers minor update would make Node.js _measurably_ slower, which many would treat as being more important than «some unintersting stuff with Buffers», and that would delay the updates even further.

And don't forget that this danger would be invisible to everyone who is running and old version of Node.js before they actually get pwned.

Do you see the problem now?

joepie91:
I just want to add that not only does Rust implement an intentionally scary-sounding 'unsafe block', [React does something similar](https://facebook.github.io/react/tips/dangerously-set-inner-html.html) with their dangerouslySetInnerHTML, for precisely these reasons.

trevnorris:
@joepie91 

> This is an ancient excuse for insecure-by-default code, and it does not help anybody. It just harms people by increasing the likelihood of vulnerabilities, whether due to mistakes or lack of knowledge.

The argument is to clearly demonstrate that what you, I and others consider "insecure-by-default" has different meaning. Which is also my argument against using "unsafe" in the name. It is not _technically_ what's happening. It's what you _perceive_ is technically happening. 

@jasnell I'm not opposed to the new signature, but I wonder whether it should migrate for consistency sake. I'll leave that up for discussion. As far as the change, would just need to rename the file to `XXX-<file name>`. I say this simply because this will probably land before the currently labelled 002 lands. But if it's a hassle then I can change it before landing.

joepie91:
> The argument is to clearly demonstrate that what you, I and others consider "insecure-by-default" has different meaning. Which is also my argument against using "unsafe" in the name. It is not technically what's happening. It's what you perceive is technically happening.

No, it's not. It is objectively, undisputably what happens from a technical point of view. You allocate a Buffer that does not come with the usual/expected memory safety guarantees, rendering it 'unsafe'. Again, what you do with it afterwards is irrelevant. Again, you are not making a technical argument.

EDIT: [Another reference.](http://blog.codinghorror.com/falling-into-the-pit-of-success/)

wbl:
Yes. Buffer(number) will now be safe on new versions, and not old ones. That's what security fixes do.

Your argument is that by waiting for the depreciation and new API's to take effect we gain because developers will in the short term switch to new APIs, while if we also fix the constructor they won't. Of course, the depreciation and new APIs will also not exist on old versions, will also take time to work themselves down into distributions etc, and perhaps even worse, will lead to code breaking on old versions. Developers who have other things on their mind will take a while to switch. Meanwhile the vulnerability stays.

Making Buffer(number) safe solves the problem for code that is not as actively maintained as we would like. It makes auditors have an easier time of things, as they can just check the version instead of grepping an entire codebase.

jasnell:
@trevnorris ... sorry, not quite sure what you mean by, "but I wonder whether it should migrate for consistency sake." :-)

trevnorris:
@jasnell you mentioned `alloc(n[, fill])`, but not `allocRaw(n[, fill])`. sorry. bad use of "migrate".

pluma:
@trevnorris 

The "not technically unsafe" behaviour **has actually resulted** in security vulnerabilities **in real world scenarios** so I don't think your argument on semantics is anything other than philosophical.

In good APIs doing something dangerously should feel profoundly different from doing something the safe way. Allocating uninitialized memory creates a vulnerability beginners (and evidently many non-beginners too) would not expect in a memory "safe" language like JavaScript so it's inherently a dangerous operation.

Note that **dangerous** also means this:

> likely to cause problems or to have adverse consequences

That definition perfectly describes the act of creating a slice of uninitialized memory in a JavaScript application. Whether the vulnerability is instantly remedied by the developer in the next line of code or whether the application suffering from the vulnerability is isolated by a firewall doesn't matter.

Consider this: power tools are inherently dangerous (i.e. you have to know how to use them properly or you will risk injury or "adverse consequences"), especially when used without their safety devices (e.g. shielding protecting the user against rotating blades). That doesn't mean there is no safe way to use them in their unsafe state. It just means that you shouldn't mess with them and you expect warning labels preventing the ignorant from killing themselves.

EDIT: To quote [Atwood](http://blog.codinghorror.com/falling-into-the-pit-of-success/):

> a well-designed system makes it easy to do the right things and **annoying (but not impossible)** to do the wrong things

(emphasis added)

jasnell:
yeah, `allocRaw(size)` is still just the one argument... largely because it doesn't make sense if we're saying `allocRaw` returns uninitialized memory :-). **If** we go with the updated signature, it might make sense to throw if a second argument is passed to `allocRaw` to ensure that someone that does switch will notice the change.

mikeal:
> The "not technically unsafe" behaviour has actually resulted in security vulnerabilities in real world scenarios so I don't think your argument on semantics is anything other than philosophical.

As has been said by the reporters of those issues the lack of zero-fill only made the problem worse, it wouldn't have actually fixed the issue (they would have still had a DOS vector, just not a remote disclosure). The fix they are recommending is to have a `Buffer.from()` API which doesn't support integers.

ChALkeR:
@trevnorris 

> The argument is to clearly demonstrate that what you, I and others consider "insecure-by-default" has different meaning. Which is also my argument against using "unsafe" in the name. It is not technically what's happening. It's what you perceive is technically happening.

Technically, but that's not what matters here.

Technically I have completely no means of knowing what is happening in «real world», technically everything that I know about the world is what I percieve.Technically I am not even touching my keyboard right now, technically I never ever left the protective shell made out of bones. Technically our bodies are all [made from stardust](https://en.wikipedia.org/wiki/Nucleosynthesis), technically I even doubt that there is «now».

Marking it as «unsafe» would have a positive effect on the ecosystem security. That's what matters.

Why I'm not strongly for putting the «unsafe» word to the method name is the fact that I estimate that positive effect being quite low, given that all [the doc requirements](https://github.com/nodejs/node-eps/pull/4#issuecomment-172618465) are met.

I'm also not sure what negative effects on the ecosystem would adding the world «unsafe» have, and this is why I am waiting for what the proponents of that would tell.

mafintosh:
After reading some of the above I just wanna reiterate that **all actual vulnerabilities found in modules so far** was related to `Buffer(number)` being called instead of `Buffer(string)`. The `Buffer.from` proposal here solves that nicely.

ChALkeR:
@mafintosh 

> all actual vulnerabilities found in modules so far was related the Buffer(number) being called instead of Buffer(string)

Not all of them. Most part, but not all. There are two problems, let's fix them both while we can.

mafintosh:
@ChALkeR sorry, i meant the ones i know about.

joshmanders:
Why is this even being debated? This should be safe from the start, I as a newbie to Buffer shouldn't have to go treading through documentation, blog posts, and heaven forbid ebooks on how to use Buffer safely.

trevnorris:
@joepie91 

> No, it's not. It is objectively, undisputably what happens from a technical point of view. You allocate a Buffer that does not come with the usual/expected memory safety guarantees, rendering it 'unsafe'.

I could see this argument with the old API, but now we have a new shiny API being discussed where requesting uninitialized memory is an explicit call. So we're mitigating the "unsafe" part (as far as accidental usage by user causing them to do something stupid) by making the call separate, documented, explicit, etc.

A method named `unsafeAlloc()` directly reflects on the developer using it. You're telling them "regardless of who you are and what you're doing, you are being unsafe by using this method". This isn't correct or cool. Since it's now no longer accidentally unsafe, a name that is actually _technically_ correct as far as what's happening in the computer (not what could happen from the user) would be `allocateUninitialized()`, but that's bloody long. So `allocRaw()` is a more concise but still descriptive API name.

@pluma 

> The "not technically unsafe" behaviour has actually resulted in security vulnerabilities in real world scenarios so I don't think your argument on semantics is anything other than philosophical.

Yes, the old API facilitated this sort of sloppy accident to happen but as I said a few lines above we have ourselves a new API that requires an explicit call to receive uninitialized memory.

> That doesn't mean there is no safe way to use them in their unsafe state. It just means that you shouldn't mess with them and you expect warning labels preventing the ignorant from killing themselves.

Precisely. We're giving the new API an explicit API. A developer will have to actively do _more_ to request uninitialized memory.

jasnell:
@joshmanders ... I believe there's consensus at this point that a change is required, right now we're simply trying to narrow in on exactly what the change should be. At the same time, there are LTS concerns that also need to be addressed given how extensively `Buffer(number)` is used so we need to make sure that the change that is made is the right one.

joepie91:
> I could see this argument with the old API, but now we have a new shiny API being discussed where requesting uninitialized memory is an explicit call. So we're mitigating the "unsafe" part (as far as accidental usage by user causing them to do something stupid) by making the call separate, documented, explicit, etc.

Again, I am not arguing 'unsafe' in the context of how it's used. I'm arguing 'unsafe' in terms of memory safety, which it still is, regardless of whether it's separated out as a method.

> A method named unsafeAlloc() directly reflects on the developer using it. You're telling them "regardless of who you are and what you're doing, you are being unsafe by using this method". This isn't correct or cool.

Okay, I'm just going to be extremely straightforward at this point: this is complete and utter nonsense. This is a _technical_ issue on a _technical_ project, and from a _technical_ point of view, the function call that is being made is not memory-safe. The function name reflects that fact.

To repeat: this is a _technical_ issue that requires a _technical_ solution. I am frankly at a loss for words that you would argue against safeguards because "the developer might feel offended" (paraphrased), and I would say that it is _extremely_ worrying if the 'unsafe' naming were to be rejected on this basis alone.

You are prioritizing perception of harm (namely, personal offense) over actual harm being caused, and that is _not okay_. **This is dangerous functionality, that can cause actual harm to real people when not used extremely carefully, and needs to be marked as such.**

I really can't believe I'm even having to have this discussion.

pluma:
@trevnorris 

You seem to be mostly agreeing with me but I would uphold that "allocRaw" doesn't ring any alarm bells. Even "allocateUnitialized" isn't obvious dangerous unless you know the perils of uninitialized memory (and I don't believe a random novice reading a "performance tip" about using that method over the safe alternative on a dodgy website would).

We're talking about creating two "explicit" APIs that are superficially similar but have drastically different security implications. There are two options:
1. Make the use of uninitialized memory distinctly more contrived than the safe equivalent (not just the default API, the actual explicit give-my-initialized memory API it currently mimicks).
2. Name it something that you wouldn't type without risking an "I told you so" moment when things go wrong because of it.

Option 1 is obviously out of the question, so option 2 is the only way to go.

If you think "unsafe" or "dangerous" might hurt someone's feelings, [I hope you never work on React](https://github.com/facebook/react/blob/0.14-stable/src/React.js#L67).

trevnorris:
@joepie91 I did foresee latching onto that phrase after posting, and unfortunate I wasn't being more explicit about it. It was to demonstrate that the word "unsafe" has connotation outside of what you're defining it to be. Whereas an API like `allocateUninitialized()` is clear and has no ulterior meaning. Sorry to have gotten you so worked up.

joepie91:
@trevnorris Well, I'm happy to hear that it wasn't meant quite as I'd interpreted it, but it still worries me. The _intention_ here is to ring alarm bells - as that is precisely what makes people think more carefully about using it - and I haven't seen any alternative proposals yet that have the same effect and are still technically accurate.

If there are alternative options that satisfy the above criteria, then I'd be happy to consider them as alternatives - but if there aren't, then I really don't feel that possible connotations should trump security here.

glenjamin:
> A method named unsafeAlloc() directly reflects on the developer using it. You're telling them "regardless of who you are and what you're doing, you are being unsafe by using this method". This isn't correct or cool.

I can see how this would be the reaction, when the intention is slightly different.

I think this is similar in principle to handling potential XSS vulnerabilities. Any user input is effectively _tainted_, until something is done to clean it up - and from that point on it's safe to use.

The idea with the `unsafe` API, is to indicate that the return value is unsafe until otherwise treated - not that the function call itself is unsafe. However, as you say, this isn't necessarily apparent from the name.

I did a bit of googling to find some other related work / topics
- https://en.wikipedia.org/wiki/Taint_checking
- https://dzone.com/articles/understanding-sunmiscunsafe
- https://developers.google.com/java-dev-tools/codepro/doc/features/audit/audit_rules_com.instantiations.assist.eclipse.auditGroup.taintedUserInput?hl=en
- https://golang.org/pkg/unsafe/

Go and Java, whose VMs operate at a similar level of abstraction to JS, both go with unsafe - so there's certainly a precedent for that terminology (not that this is necessarily in itself a compelling argument).

Personally, I think the longer names like `allocUninitialized` are pretty good, I would expect that Node.js programs won't generally be filled with allocations of raw memory (except for core).

Some potential candidates, including names already suggested (yes, the thesaurus was involved)
- `allocUninitialized`
- `allocRaw`
- `allocRawMemory`
- `allocUnsafe`
- `allocUnsafeMemory`
- `allocTainted`
- `allocDirty`
- `allocPotentiallyUnsafe`
- `allocDangerous`
- `allocInsecure`
- `allocCrude`
- `unsafeAlloc`
- `allocRisky`
- `allocUncertain`
- `allocPrecarious`
- `allocOrganic`
- `allocInfected`
- `allocMemoryYouNeedToFillToAvoidMemoryDisclosure`

Of these, I quite like `allocDirty` - there's an implication it can be cleaned.

trevnorris:
@pluma 

> Even "allocateUnitialized" isn't obvious dangerous unless you know the perils of uninitialized memory (and I don't believe a random novice reading a "performance tip" about using that method over the safe alternative on a dodgy website would).

I can't accept that about our community of developers. If someone is writing software for the server (especially if they plan to publicly expose it) and manually allocate memory they should understand the difference between initialized and uninitialized memory. Do you want to put bright blinking exclamation marks in the documentation explaining that it's not only different but also unsafe? Sure, be my guest.

> If you think "unsafe" or "dangerous" might hurt someone's feelings, I hope you never work on React.

Haha. Epic. That made my day.

@joepie91 Saying that the "_intention_ here is to ring alarm bells" you've also demonstrated wanting to change the developers emotional state simply by using the API name. What I surmise is that to you the API name has to first "ring alarm bells" and only after that must it be "technically accurate".

Just reflecting back on my own experience and that of other developers I worked along side. It's most probable that the group of developers we're helping guard against themselves the most are those from the client-side (I say this having come from the same). Problem is looking back I don't ever recall an API name or such preventing us from trying something out. It more depended on the examples and articles found online. That said, I feel like the more important step to take after this is released is a push to get new/updated content on the web and help projects to start using the new API. By example we learn, right?

ChALkeR:
@pluma 

> and I don't believe a random novice reading a "performance tip" about using that method over the safe alternative on a dodgy website would

You just increased my estimation of the positive effect from naming it «unsafe» (more precisely, of the negative effect from not doing so).

glenjamin:
> If someone is writing software for the server (especially if they plan to publicly expose it) and manually allocate memory they should understand the difference between initialized and uninitialized memory.

Personally, I'd never really thought of buffer this way. When I create an array or a string I'm not thinking about memory layout or even really about allocation, I just want some stuff. The GC takes care of the rest until I need to optimise a hot path.

Normal buffer usage is when I want a sequence of bytes I can send to a stream. In this respect it's not dissimilar to a mutable array - except the file and network APIs will accept buffers, but not arrays. One of the main reasons to program network stuff in Node instead of the more traditional C, is to avoid having to think about unsafe memory access and other low-level concerns.

I reckon allocating an uninitialised buffer is a power-user optimisation to speed up hot paths, not necessarily something people should need to reach for immediately. As such, making it slightly scarily named might put people off using it until they need to (and have read up on it). This might be an ok choice from an API design perspective.

joepie91:
> a random novice reading a "performance tip" about using that method over the safe alternative

This is precisely the kind of thing I'm worried about.

> I can't accept that about our community of developers. If someone is writing software for the server (especially if they plan to publicly expose it) and manually allocate memory they should understand the difference between initialized and uninitialized memory.

It is not obvious from Buffer that it is "allocating memory". To the average developer, it looks like a way to define a safe encoding-less "byte array". This is unfortunate, but realistic.

> Saying that the "intention here is to ring alarm bells" you've also demonstrated wanting to change the developers emotional state simply by using the API name.

Yes. This is an inherent part of UX design, in all of its forms. Affecting emotional state is how certain behaviour is driven, in one way or another.

(EDIT: Just to clarify, I'm not a fan of manipulation in any sense, and I frequently campaign against it, in many contexts. My first priority here, however, is to prevent immediate harm from being caused - people _are_ being manipulated on a daily basis, and because of eg. the influence of marketing in how important people perceive 'performance' to be, there is a real risk here.)

> What I surmise is that to you the API name has to first "ring alarm bells" and only after that must it be "technically accurate".

No, both are equally important, as far as I'm concerned.

> Just reflecting back on my own experience and that of other developers I worked along side. It's most probable that the group of developers we're helping guard against themselves the most are those from the client-side (I say this having come from the same). Problem is looking back I don't ever recall an API name or such preventing us from trying something out. It more depended on the examples and articles found online.

There are quite a few concrete examples of cases where the "pit of success" design has considerably reduced issues. See for example the design of libsodium (which almost nobody uses or documents incorrectly), or the usage of Rust versus C. I am not aware of any cases that show the opposite.

These "examples and articles" you are referring to started somewhere - more often than not, they are the result of somebody needing something to write about, without investing too much time into researching the topic (or even reading the documentation). These people, too, would be steered away from a function named "unsafe".

> Some potential candidates, including names already suggested (yes, the thesaurus was involved)

`allocDirty` seems the most reasonable to me from a 'technical accuracy' and 'alarm bells' point of view (alongside `allocUnsafe`), but I do have to say that it sounds a little strange at first. That may just be my perception, though.

jasnell:
I'm going to call time on this part of the discussion -- debating back and forth about whether allocating uninitialized memory is "safe" or not is not productive right now. Everyone agrees that using uninitialized memory is not a good thing, debating about semantics is not. 

There are three fundamental problems to be addressed:
1. The current API design makes it possible to accidentally use `new Buffer()` in unsafe ways.
2. `new Buffer(number)` allocates uninitialized memory by default
3. `new Buffer(number)` is used so extensively by the ecosystem that we absolutely **must** be sensitive to introducing potentially breaking changes. _Forcing_ performance slowdowns in production code by changing the implementation **CAN** break existing code depending on the specific requirements of that code. We need to be deliberate about the changes that are made.

To be absolutely clear, the proposal I'm currently putting on the table is this:

A. To address (1), we:
1. introduce separate `Buffer.alloc()`, `Buffer.allocRaw()`, and `Buffer.from()` methods with improved argument type checking and;
2. deprecate in _docs_ (see point C below) `new Buffer()` explaining exactly why developers should not use `new Buffer()` going forward.

B. To address (2), we introduce the `--zero-fill-buffers` command line argument to master, v5, v4, v0.12 and v0.10 to force all `Buffer` instances to zero fill. This change would allow anyone who is concerned around the security of using uninitialized Buffers to force all code, whether maintained or not, to zero-fill every buffer.

C. To address (3), we avoid hard deprecation of `new Buffer()` in v5, v4, v0.12 and v0.10 but make it very clear via release notes, documentation and blog posts that `new Buffer()` is dangerous and that implementations should either (a) move to new versions of Node and use the new APIs, (b) use the `--zero-fill-buffers` command line argument, or (c) polyfill the new APIs. 

This proposal currently addresses most of the scenarios save one: Developers who _don't know_ that the modules they are using are not using `new Buffer(number)` correctly and who _don't know_ that they should use `--zero-fill-buffers` as a backstop.  For such users, switching the default behavior of `new Buffer(number)` to zero fill always _is the easiest solution_ but, as item 3 above indicates, there is a non-zero risk of breaking a significant amount of production code if such a decision is made without care. I am currently exploring the options to address this concern. I've narrowed it down to two fundamental choices:

A. We stick with the current proposed solution of maintaining the current behavior and adding the `--zero-fill-buffers` -- in which case, we have the burden of educating everyone that they need to update.

B. We switch the default behavior and add an `--unitialized-buffers` command line argument instead to allow those who depend on the current behavior to keep functioning as they currently do. We then have the burden of educating those users and explaining why they need to update their environments to use the new flag. 

As a point in reference, there is precedence for changing defaults in order to give Node a more secure stance by default. Specifically, when we updated the default TLS cipher list to disable things like RC4 by default we made the change in v0.10 and v0.12 also. This change also broke production code in some non-trivial ways, but overall was the appropriate change to make.

joepie91:
@jasnell This is not "debating about semantics". There are significant, real-world security consequences in the choice of naming of a dangerous API. This is a discussion that needs to be had.

Several technical arguments have been raised for `allocUnsafe`, which as of yet have not been invalidated. I am not aware of any standing technical arguments against `allocUnsafe` and for `allocRaw`. If we're going to default to _something_, I can only see `allocUnsafe` as being the reasonable option.

Aside from that, I agree with A and B (from the first list). I am on the fence about changing default behaviour in older versions.

trevnorris:
Had some sushi and wondered around the internet on usage of "unsafe" in in other languages, etc., for uninitialized memory allocations. Rust took the cake (and almost all my Google search results). Mainly the function `std::mem::uninitialized`, which is `unsafe`. Also there was Java's `sun.misc.Unsafe` to allocated uninitialized Objects.

Anyway. I still question the definition of "technically correct" when it comes to using "unsafe" in this case, but `allocUnsafe()` seems reasonable enough. Though I still prefer `allocRaw()`.

/cc @bnoordhuis any thoughts on this matter?

trevnorris:
@jasnell Nice write up. Agree with everything there, and if investigation (whatever that would entail) shows low likelihood of breakage, swapping implementation to zero-fill by default seems reasonable. But being able to quantify that appropriately makes this case almost impossible.

ChALkeR:
@jasnell As I have stated multiple times, including here, I'm strongly against switching the default behavior and adding an optout, it will have negative security consequences. People who write new libs and people who already use `.fill(0)` will not use `.fill(0)` anymore to prevent double zero fills (many won't see a point in that). If one tells them that now their code is broken with a flag, they would tell «don't use that flag». Moreover, the optout will effectively work as a kill switch. I view this not as a backwards compatibility problem (like it was with TLS cipher list), but as an actual issue that will _negatively affect the security_.

Everything else looks good to me. Though I would prefer a method name that triggers more concern when one decides to use it over the method name that triggers less concern.

ChALkeR:
@jasnell Regarding the opt-in, I initially proposed a different flag that would zero-fill only buffers allocated by `Buffer(number)` and that would be removed when `Buffer(number)` would be hard-deprecated (whenever that is). Let's call it `--zero-fill-buffer-number` for now.

Those flags have different effect, usecase, and do not replace one another:
1. `--zero-fill-buffer-number` is needed if one trusts the new libs to be accurately using the new `Buffer.alloc()/Buffer.allocUnsafe()/Buffer.from()` API, but protects against the legacy code that was using the old `Buffer()` API. There should be no measurable performance degradation here in most cases (core would be updated to use the new API), this mostly a sane thing that everyone should turn on, it's an opt in to prevent the problems that I described and to give users a grace period without causing more security issues. It could be removed when `Buffer(number)` is hard-deprecated and hardcoded in an «enabled» state. This is a temporary flag.
2. `--zero-fill-buffers` is needed when one does not care about small (probably) but measurable performance penalties to be more secure, and does not trust libraries to be accurately using the new `Buffer.alloc()/Buffer.allocUnsafe()/Buffer.from()` API. This is a long-term flag that would not be removed in the future. This one is probably needed regardless of other things, and is a valuable positive change by itself.

Though, I think that adding _two_ of them is perhaps excessive. Or not?
Opinions?

jasnell:
I believe it is definitely excessive and would lead to more confusion. For those that need `--zero-fill-buffers`, the need for ensuring initialized buffers is going to outweigh the need for performance. It should be something that they can turn on and rely on without worrying about which modules they can trust and which they may not be able to.

feross:
If folks want a userland module that implements the proposal, there is: https://www.npmjs.com/package/safe-buffer

Once the new APIs are in core, I'll update `safe-buffer` to use the built-in implementations when available, but still preserve support back to node 0.10.

ChALkeR:
@feross I believe that there is still no final consensus on the method names, so it might be too early to provide that.

jasnell:
FWIW, the overall tide in the discussion seems to be leaning towards either favoring or being ok with `allocUnsafe` as the naming. While I recognize that not everyone agrees with that naming, I would wager that a majority could likely live with it.

The only remaining controversial points would appear to be (a) whether or not we switch the default behavior of `new Buffer(number)` to zero-fill and (b) if we backported such a change to LTS streams. The key deciding point on that _appears_ to be just how much of an impact such a change would have on the ecosystem. If anyone has any ideas on how we can begin to quantify that, I'd be thankful :-)

feross:
@ChALkeR It's in userland, so it's fine. We'll bump the major version once this lands in node core and use the finalized method names.

ChALkeR:
@feross Well, as you wish. I just don't see the point of publishing it beforehand, because whoever relies on it now would have to change his or her code pretty soon.

@jasnell Neither of those methods (`Buffer.alloc/Buffer.allocUnsafe/Buffer.from`), seems to be directly present in my (a bit old, from 2015-09-21) dataset. They still could be present in some modules that actually extend `Buffer`. About the possible consequences that are not related to the current code but are related to the user actions after the changes are made — I already expressed my concerns, but perhaps some Node.js versions distrubution stats would give more information here.

ChALkeR:
@jasnell Also note that `Buffer.from()` is already present, so this proposal will override that. But it doesn't look like it is (or was on 2015-09) used.

Are there any situations when the old `Buffer.from` would accept input and return different results from the new `Bufer.from`?

To me, it seems that the old `Buffer.from()` just always throws.

jasnell:
@ChALkeR ... thank you for that info. Perhaps the more difficult part to measure is how extensive of an impact changing the default behavior of `new Buffer(number)` would have.

trevnorris:
@ChAlKeR The existing `Buffer.from()` is just `Uint8Array.from()`. We'll be overriding at least half the ways it handles input. Most commonly would be how strings are handled.

feross:
I just found a few more modules affected by this issue:
- https://github.com/request/request/pull/2018
- https://github.com/rvagg/bl/pull/22

Not a huge deal, since I think they're pretty hard to exploit. But wanted to share the additional data points.

joepie91:
@jasnell I've thought about the backporting strategy a bit more, overnight. I can see the following options:
1. **We backport automatic zero-filling, back all the way to 0.10.** This would likely be a justifiable change (given [your earlier explanation](https://github.com/nodejs/node-eps/pull/4#issuecomment-172688180)), and would provide immediate protection for everybody using a supported version, _as long as they update_.
   
   The upside is that it does not introduce additional methods in older versions. The downside is that _it fails deadly_ when people use an outdated version, as it will silently fail to zero-fill while userland module developers believe that it'll zero-fill correctly.
2. **We backport `alloc`, `allocUnsafe` and `from`, back all the way to 0.10.** This would add API surface - however, per [@ChALkeR's data](https://github.com/nodejs/node-eps/pull/4#issuecomment-172707412), this should not cause any conflicts (and further, the recommendation to not modify prototypes you don't own has been around for a very long time, for precisely this kind of reason).
   
   This _fails safe_, in that if people use an outdated version of Node, it will simply throw a hard error when these methods are used, instead of quietly doing the wrong thing.
3. **We don't backport either of these, or we only do it for some of the newer versions.** As explained before, this will likely cause adoption problems. Many developers in the Node.js ecosystem (unfortunately) still have an aversion towards third-party dependencies and 'bloat', so a userland implementation or polyfill would likely not work very well here - rather than defaulting to the safe thing, you'll have to convince every single individual developer to care about it. This seems like the least workable solution to me.

All of the above options assume that an opt-in flag exists (and is backported) to forcibly zero-fill all buffers, application-wide. I think we're all in agreement about that part.

@ChALkeR You say that `Buffer.from()` already exists - but this is not documented, at least not in the Node.js documentation. I think it would be safe to override it, for that reason?

ChALkeR:
@jasnell 

> Perhaps the more difficult part to measure is how extensive of an impact changing the default behavior of new Buffer(number) would have.

About 8% of npm modules use `Buffer()` directly. About 2% of npm modules use `Buffer(constant number)` (e.g. `Buffer(10)`) directly. It would be harder to count how many people actualy zero-fill manually allocated buffers (because that's often done later in the code, sometimes even manually using a loop and not through `.fill(0)`), but from what I have seen, a considerable number of packages does that.

I don't think that we can numerically estimate the potential damage introduced by people removing the zero-fill.

trevnorris:
@jasnell Note on why `Buffer.from()` can't be used: https://github.com/nodejs/node/pull/4682#issuecomment-172983765

ChALkeR:
`Buffer.value()`, `Buffer.fromValue()`, `Buffer.create()`, `Buffer.val()`, `Buffer.make()`, `Buffer.toBuffer()`, `Buffer.init()`.

jasnell:
I've got Buffer.from working correctly in the PR with no issues. All tests
passing.

trevnorris:
Is `Buffer()` eventually meant to not be a constructor? Don't think that's really possible.

lime:
> @joepie91: **We backport automatic zero-filling, back all the way to 0.10.** [...] The downside is that it _fails deadly_ when people use an outdated version, as it will silently fail to zero-fill while userland module developers believe that it'll zero-fill correctly.

It seems like the most significant argument against a backported change to `new Buffer(10)` is this, that people will start depending on its new, predictable behaviour. I'm inclined to believe that this could likely happen.

However, the conclusion drawn from this has been to paint any backported change as dangerous. I think this can be revisited.

-- A silly proposal --

Consider this: the only requirement for a less-dangerous `new Buffer(10)` is that it is initialized with something. At no point was it a requirement that it is initialized with _zeroes_. It can be anything at all, and as long as it's not your private keys, we should be fine.

This was touched upon in https://github.com/nodejs/node/issues/4660#issuecomment-171764113:

> The existing API specifies that the memory is uninitialized, which means it can have any value. Zeros are a valid value.

Zeroes are indeed a valid value. So are ones. So are the bytes in the ASCII string "banana", repeated over and over. Or the current Node.js version string, followed by all zeroes. You name it.

Now don't get me wrong, zeroing out is absolutely the most sensible choice **in every other way**. It's backwards compatible, in the sense that all-zeroes was always a valid outcome with uninitialized memory. It's much more performant than repeating an ASCII string, or some other silly initialization. The **only** drawback is that it creates a new pitfall, making people wrongly assume that zeroing out happens in all versions.

If (and only if) that pitfall is determined to be a deal breaker, _I propose that we initialize `new Buffer(10)` with something other than zeroes._ This affects only the constructor, `.alloc()` and `.allocUnsafe()` do not change.

-- Why that's undesirable --
1. Performance: at multiple occasions, there have been people proposing that nothing at all should be backported because of performance concerns. Seeing as zeroing with `calloc` is very likely the most performant initialization we can do, any other initialization would be difficult to justify.
2. Determinism: just as people would start depending on the all-zeroes implementation, there will be a non-zero number of developers who (ab)use any deterministic initialization. "Need to write 'banana'? Just do `new Buffer(6).toString()`." This is largely a non-issue though, as long as it's different from `.alloc()`.

-- The tradeoff --

Is the depending-on-all-zeroes pitfall so great that we can sacrifice some more performance in initialization?

ChALkeR:
@trevnorris 

> Is `Buffer()` eventually meant to not be a constructor? Don't think that's really possible.

But it could throw on all values. Eventually.

ChALkeR:
@lime 

> I propose that we initialize new Buffer(10) with something other than zeroes. 

That makes no actual difference with being zero-filled.

> Is the depending-on-all-zeroes pitfall so great that we can sacrifice some more performance in initialization?

I do not view this (zero-fill-all-the-buffers) approach to be a performance vs security tradeoff. I view it as a bad ecosystem security vs even worse ecosystem security tradeoff.

When I mention performance concerns, I mention that those are the concerns that _some_ people would have when updating to new Node.js (hence delaying update and staying on the unpatched version) and _other_ people would have when deciding whether or not keep their own initialization of allocated `Buffer`s (e.g. `.fill(0)`, `.fill(42)`, for loops, or whatever — I saw various types of those) on top of the default one in the updated Node.js. Combine that and it goes boom.

Also, the proponents of just-zero-fill-all-the-buffer-and-dont-touch-the-api approach are forgetting that it alone would not even fix the issue completely even in the long term (taking only patched versions of Node.js into an account).

jasnell:
Proposal PR has been updated to use the names `alloc()` and `allocUnsafe()`

jasnell:
Thinking further about the issue over the default behavior of `Buffer(number)` moving forward. While the deprecation of the constructor would be doc only (i.e. "soft deprecation") one concrete step we can take to begin educating users to move away from use of the constructor is print a one time warning to stderr similarly to how we print the "possible memory leak" warning when adding more than 10 listeners to an EventEmitter event. This would at least give us the ability to notify users that something is potentially amiss and should be looked at.

joepie91:
@lime: The issue is not so much that people would expect a _zero-filled_ Buffer, but that they would expect a _safely filled_ Buffer - ie. not containing stuff that was previously in memory. That problem is not resolved by filling the Buffer with non-zero (or even random) data, unfortunately, and the same consequences remain applicable.

@jasnell, @trevnorris: How likely is it that `.from` will break (ie. tests not passing and no real way to fix it) with V8 updates? This sounds a bit hacky, although I don't really know enough about V8 internals to say that with any certainty.

Aside from that:
- `Buffer.fromValue`, while a little longer, still sounds like a good alternative name to me, if it were needed.
- `Buffer.value` and `Buffer.val` might create confusion for those who have come to expect a getter/setter function from something named that (eg. jQuery users).
- `Buffer.toBuffer` sounds confusing to me (are we creating a Buffer from a Buffer? Huh?)
- `Buffer.make`, `Buffer.create` and `Buffer.init` don't sound _bad_ to me, but their purpose may not be obvious from reading the code.

> While the deprecation of the constructor would be doc only (i.e. "soft deprecation") one concrete step we can take to begin educating users to move away from use of the constructor is print a one time warning to stderr similarly to how we print the "possible memory leak" warning when adding more than 10 listeners to an EventEmitter event. This would at least give us the ability to notify users that something is potentially amiss and should be looked at.

While I like this idea, we should expect some backlash from some users who don't like warnings to appear in their `stderr`. I've seen some people complain about this for deprecation warnings in other projects in the past. I _personally_ don't feel that it's a major problem, though.

lime:
@joepie91 & @ChALkeR: Since my proposal is slightly tongue-in-cheek, I won't make any big attempts to defend it. However, I'd like to try clarifying my intent just a bit.

The way I see it, the **documented** behaviour could go something like this:
- `alloc()`: Allocates a zero-initialized buffer.
- `allocUnsafe()`: Allocates an uninitialized buffer.
- `new Buffer(number)`: DEPRECATED. Allocates a buffer, making **no guarantees** about memory safety.

Now, the problem at hand is that documentation isn't enough. Instead of reading / remembering this documented behaviour, people will take a look at the **observable behaviour** of `new Buffer(number)` on their version, and assume it applies for all versions.

If the observable behaviour of `new Buffer(number)` is different from `alloc()` in whatever way, it's ever so slightly harder to make dangerous assumptions based on that behaviour.

lime:
The suggested deprecation warning printed to `stderr` would do a lot to alleviate these concerns. It's much harder to make dangerous assumptions from observable behaviour, when that behaviour includes a written warning. :)

ChALkeR:
@lime 

> `new Buffer(number)`: DEPRECATED. Allocates a buffer, making no guarantees about memory safety.

That would work only in an ideal world where everyone reads docs and follows them. But, as I already said above, it should be good to zero-fill `Buffer(number)` once that is **hard**-deprecated.

ChALkeR:
> The suggested deprecation warning printed to stderr would do a lot to alleviate these concerns. It's much harder to make dangerous assumptions from observable behaviour, when that behaviour includes a written warning against such assumptions. :)

Given how many of modules are currently using `Buffer()`, and how it would not be possible to simply switch to the new API without losing compatibility with the old Node.js versions (including, for example, 4.2.x and 5.4.1) or adding a feature-detection hack, I doubt that that warning will do anything good in the nearest future, it will likely only annoy people.

lime:
> That would work only in an ideal world where everyone reads docs and follows them.

I believe I specifically said, in the comment you quoted, that documentation isn't enough. That is why I suggest that `new Buffer(number)` should do _something_ to differentiate it's observable behaviour from `alloc()`.

That _something_ can be a different kind of initialization. FWIW, I don't think this would be desirable.

That _something_ can also be a deprecation warning, as suggested by @jasnell. Personally, I think this sounds like the best approach.

trevnorris:
@ChAlKeR Buffer at minimum must be the Uint8Array constructor. ES6 inheritance requires this. So throwing on all values also isn't an option.

rvagg:
For completeness we need to talk about the native `node::Buffer` API and whether any changes are in scope or not and be explicit why not if they are not on the table. We need to be careful to set expectations properly if they are going to operate differently.

jasnell:
@rvagg ... Yep, I was going to turn my attention to the native side next
once it was clear we had things fairly well settled on the js side.

jasnell:
Closed this because I realized I had screwed the pull request up. Opening a new corrected PR. Discussion can continue here, of course.

jasnell:
Refs: https://github.com/nodejs/node-eps/pull/7

PR #5, opened 2016-01-23
fix markdown link for Public C++ Streams

targos:


trevnorris:
Thanks much. Landed in 817370bb98.

PR #6, opened 2016-01-25
ep: ws in core

eljefedelrodeodeljefe:
EP as on nodejs/node#4842

### DIFF
--- a/003-ws-in-core.md
+++ b/003-ws-in-core.md
@@ -1,0 +1,30 @@
+| Title  | ws in core                  |
+|--------|-----------------------------|
+| Author | @eljefedelrodeodeljefe      |
+| Status | DRAFT                       |
+| Date   | 2015-01-25 02:34:00         |
+
+## Description
+
+This EP proposes having the current [websockets/ws](https://github.com/websockets/ws) userland implementation in core.
+Efforts have already been made on nodejs/node#4842. It is understood that
+adaptions to style and simplicity need to be done in order to conform with
+`node` core.
+
+## Motivation
+
+> `websockets` are the primary reason for companies, which have **not** decided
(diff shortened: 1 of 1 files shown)


003-ws-in-core.md
```diff
@@ -0,0 +1,30 @@
...
+This EP proposes having the current [websockets/ws](https://github.com/websockets/ws) userland implementation in core.
+Efforts have already been made on nodejs/node#4842. It is understood that
+adaptions to style and simplicity need to be done in order to conform with
+`node` core.
+
+## Motivation
+
+> `websockets` are the primary reason for companies, which have **not** decided
+to make a commitment towards `node` to use `node` anyways. For a wider adoption
+of the language this is a critical point.
```
Fishrock123: How much basis is there for this? My insight indicates to me: little to none..

eljefedelrodeodeljefe: ...me as an engineering manager in a PHP-shop.

eljefedelrodeodeljefe: Could change wording, but I think this fact still hold up.

ChALkeR: I highly doubt that this is a valid point. Without performing some kind of relevant research, at least.

feross: If companies are already adopting node.js because of how good the `ws` package is, then that means the `ws` maintainers are doing a great job and there's nothing that node core needs to do.

trevnorris:
@eljefedelrodeodeljefe Now that v8_inspector has landed, with basic ws support, may want to revise this proposal.

styfle:
The most recent discussion is happening in https://github.com/nodejs/node/issues/19308

Trott:
Superseded by https://github.com/nodejs/node/issues/19308

PR #9, opened 2016-02-12
002: Per-isolate exit hooks

dherman:
A proposal for a `node::AtIsolateExit` API for addons to reigster tear-down behavior for individual isolates. Initially this doesn't offer any benefit over `AtExit`, but when workers land, it will allow teardown behavior for each isolate, including not just the main process but each individual worker as well.

### DIFF
--- a/002-per-isolate-exit-hooks.md
+++ b/002-per-isolate-exit-hooks.md
@@ -1,0 +1,54 @@
+| Title  | Per-isolate exit hooks      |
+|--------|-----------------------------|
+| Author | @dherman                    |
+| Status | DRAFT                       |
+| Date   | 2016-02-12                  |
+
+## Description
+
+The public C++ API has the [`AtExit`](https://nodejs.org/api/addons.html#addons_void_atexit_callback_args) API for specifying teardown logic for a Node process, but with the upcoming addition of workers, there will be a need for custom teardown logic for arbitrary isolates.
+
+This is particularly useful for addons that want to lazily allocate per-isolate data on a by-need basis, and therefore do not control the actual creation of the workers. Addons like that need the ability to deallocate any associated data structures when the isolate goes away.
+
+For example, an addon can create a tagged native class with [`v8::FunctionTemplate`](https://v8docs.nodesource.com/node-5.0/d8/d83/classv8_1_1_function_template.html) that's associated with some additional class metadata. (This is in fact my situation for [https://github.com/rustbridge/neon](Neon).) Any workers that use the class will cause it to lazily create the class, its function template, and its metadata. In order to avoid a memory leak, the code that lazily allocates the metadata will also want to register a callback to deallocate that metadata when the worker terminates.
+
+## API
+
+**enum IsolateKind { Main, Worker }**
+  * **`Main`**: this isolate belongs to the main Node process.
+  * **`Worker`**: this isolate belongs to a worker.
+
+**void AtIsolateExit(isolate, callback, args)**
+  * **`isolate`**: `v8::Isolate *` - A pointer to the isolate to hook into.
+  * **`callback`**: `void (*)(v8::Isolate *, void *, IsolateKind)` - A pointer to a function to call when the isolate is exited.
+  * **`args`**: `void *` - A pointer to pass to the callback as the second argument.
+
+Registers exit hooks that run after the event loop of an isolate has ended.
+
+AtIsolateExit takes three parameters: a pointer to the isolate to hook into, a pointer to the callback function to run at exit, and a pointer to untyped context data to be passed to that callback.
+
+Callbacks are run in last-in first-out order.
+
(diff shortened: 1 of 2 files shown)


trevnorris:
Now that is one heck of a write up. Thanks for being so thorough. The reading looks good to me.

@bnoordhuis @indutny Either of you see any issues with implementing this?

petkaantonov:
btw I was supposed to do the file splitting today but instead i'm migrating from virtualbox to vmware

petkaantonov:
err commit splitting

002-per-isolate-exit-hooks.md
```diff
@@ -0,0 +1,54 @@
...
+
+For example, an addon can create a tagged native class with [`v8::FunctionTemplate`](https://v8docs.nodesource.com/node-5.0/d8/d83/classv8_1_1_function_template.html) that's associated with some additional class metadata. (This is in fact my situation for [https://github.com/rustbridge/neon](Neon).) Any workers that use the class will cause it to lazily create the class, its function template, and its metadata. In order to avoid a memory leak, the code that lazily allocates the metadata will also want to register a callback to deallocate that metadata when the worker terminates.
+
+## API
+
+**enum IsolateKind { Main, Worker }**
+  * **`Main`**: this isolate belongs to the main Node process.
+  * **`Worker`**: this isolate belongs to a worker.
+
+**void AtIsolateExit(isolate, callback, args)**
```
bnoordhuis, line 21: We could keep the `AtExit` name and depend on overloading.

002-per-isolate-exit-hooks.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Alternatives
+
+I'm not particularly attached to the name, but I tried to pick something consistent with the `AtExit` name.
+
+The isolate doesn't _have_ to be passed to the callback, but it seems like a nice convenience so that it doesn't have to be added to the context data structure.
+
+The isolate kind could just be a boolean `is_main` flag, but I tried to avoid [the Boolean trap](http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html) of API design.
+
+We could try to raise the level of abstraction to `node::Environment`, in order to avoid the V8-specific API. I understand there are people interested in revamping the Node C++ architecture away from V8 specifics. I think that's a great idea, but given how pervasive isolates are in the existing API, it would be hard (or maybe impossible) to make the incremental change just here. I think the more straightforward path is to add this small API now and orthogonally allow people to investigate the overall strategy for migrating Node and native addons towards a more abstract API layer.
```
bnoordhuis, line 50: I'd leave it at `v8::Isolate` for now.  The state of JS engine API abstraction is currently best described as 'research project'.

bnoordhuis:
LGTM, seems like an uncontroversial change to me; it's basically how I would design it.

@trevnorris Do we have guidelines on line length?

002-per-isolate-exit-hooks.md
rvagg, line 21: @bnoordhuis how would that look? just move `isolate` to the end and put `void*` at the start of the callback signature? +1 if we can achieve overloading here.

rvagg:
Isn't there some way we could determine the `Isolate` kind simply by having the `Isolate`? And why is that knowledge required anyway?

@petkaantonov are you able to review this and give sign-off too since it impacts the workers PR?

petkaantonov:
Well, the `Environment` already has methods to see if it's the main environment or a worker environment. IIRC you can also get the v8 isolate from the env. So maybe the isolate should be replaced with env and the Main, Worker enum is not needed?

002-per-isolate-exit-hooks.md
bnoordhuis, line 21: E.g.:

``` c++
void AtExit(void (*)(void*), void*);  // non-isolate version
void AtExit(v8::Isolate*, void (*)(v8::Isolate*, void*), void*);  // or:
void AtExit(void (*)(void*, v8::Isolate*), void*, v8::Isolate*);
```

The last one has the property that you can `reinterpret_cast` an existing callback if you don't care about the isolate parameter.

bnoordhuis:
`Environment` isn't part of the stable API however.

petkaantonov:
But this api is for the workers which are not exactly stable either. By the time workers become stable I would think Environment is stable too or?

bnoordhuis:
`Environment` is never stable, it's just a grab bag for per-context state.  It changes frequently and I don't see that changing anytime soon.

petkaantonov:
In that case LGTM

trevnorris:
@bnoordhuis 

> Do we have guidelines on line length?

Not explicitly stated, but these should follow RFC guidelines and should be no more than 78 characters. I'll add that to the README

Trott:
Closing as this is over two years dormant. By all means, comment or re-open if this is still an ongoing concern, although I suspect it should be moved to a more active repository (probably the main node repo) if that's the case.

PR #16, opened 2016-04-26
Fix wording on Status field

trevnorris:
Conflicting text was added that contradicts what the Status field can be
when an EP lands. Instead refer back to the main text where this is
explained.

Fix minor grammar mishap, change "a" to "an".

Fixes: f5875b4 "Revise Format section"

R=@jasnell 
R=@rvagg


jasnell:
LGTM

trevnorris:
Thanks much. Landed in 59aacb7d.

PR #17, opened 2016-04-27
copyedit 002-es6-modules.md

Trott:
Fix a handful typographical/grammatical issues. One word was changed
(`nonplussed` means "surprised and confused", not "unenthusiastic" which
seems closer to what was meant in the document).

I'm not sure what to do with this sentence, but it definitely needs some
attention:

``` text
If foo was providing this file explicitly like `lodash` has this can be
mitigated easily by using as proxy module should `foo` choose to provide
one:
```

### DIFF
--- a/002-es6-modules.md
+++ b/002-es6-modules.md
@@ -4,9 +4,9 @@
 | Status | DRAFT                       |
 | Date   | 2016-01-07                  |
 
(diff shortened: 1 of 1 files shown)


trevnorris:
LGTM

trevnorris:
@Trott Sorry. Not sure how I LGTM'd it but didn't land it. Though the document this PR addresses has changed. There still issues there you want to address?

PR #20, opened 2016-04-29
Initial version of eps for ABI-Stable-Module-API

mhdawson:
First cut of the ABI-Stable-Module-API eps.  It is far
from complete but is intended to allow insight into the
current direction/status and will be updated regularly
as we progress along the investigation and can make
the API definition more complete.

### DIFF
--- a/005-ABI-Stable-Module-API.md
+++ b/005-ABI-Stable-Module-API.md
@@ -1,0 +1,127 @@
+| Title  | ABI Stable Module API                 |
+|--------|---------------------------------------|
+| Author | @mhdawson, @stefanbu, @Ianwjhalliday  |
+| Status | DRAFT                                 |
+| Date   | 2016-12-12                            |
+
+# Introduction
+
+The module ecosystem is one of the features making Node.js the fastest
+growing runtime.  An important subset of modules are those which have a
+native component. In specific cases it is important to be able to leverage native
+code in order to be able to:
+
+* use existing third party components written in C/C++
+* access physical devices, for example a serial port
+* expose functionality from the operating system not otherwise available
+* achieve higher speed than possible in native JavaScript
+
+There is an existing interface (addons) that allows you to build these modules
+as a dynamically-linked shared module that can be loaded into Node.js and
+used as an ordinary Node.js module (for more details see
+https://nodejs.org/api/addons.html)
+
+Unfortunately in the existing implementation the V8
+(the JavaScript runtime for Node.js) APIs are exposed to modules. In
+addition Google changes these APIs on a regular basis. This results in
+the following issues:
+
+* Native modules must be recompiled for each version of Node.js
+* The code within modules may need to be modified for a new version of
+  Node.js
+* It's not clear which parts of the V8 API the Node.js community believes
+  are safe/unsafe to use in terms of long term support for that use.
+* Modules written against the V8 APIs may or may not be able to work
+  with alternate JavaScript engines when/if Node.js supports them.
+
+The Native Abstractions for Node (NAN)
+project (https://github.com/nodejs/nan) can be used to limit
+the cases when modules must be updated in order to work with new versions
+of Node.js but does not address the other issues as it does not
+encapsulate the V8 argument types such that modules still need
+to be recompiled for new Node.js versions.
+
+# Proposed Action
+
+The goal of this eps is to define an API that:
+
+* Provides a stable API which is not tied to the current
+  or further APIs provided by V8
+* Allows native modules built as shared libraries to be used
+  with different versions of Node.js
+* Makes it clear which APIs can use with the expectation
+  of not being affected by changes in the JavaScript engine
+  or Node.js release
+* The API provided by the Node.js runtime will be in C. In
+  addition a C++ wrapper which only contains inline functions
+  and uses the C APIs so that the changes in the C++ wrapper
+  would not require recompilation.
+
+# Background
+
+There have been ongoing discussions on a new module API in both the
+community api working group (https://github.com/nodejs/api), other
+issues in the Node.js  issue tracker, and the recent community Node.js
+face-to-face VM summit (https://github.com/jasnell/vm-summit).
+
+In that summit there was agreement that we wanted to define this
+module API.  The general approach was to be:
+
+* Start with NAN, trim down to minimal subset based on module usage
+  and inspection.
+* Add in VM agnostic methods for object lifecycle management.
+* Add in VM agnostic exception management
+* Define C/C++ types for the API which are independent from V8
+  (NAN currently uses the V8 types).
+* Use the NAN examples to work through the API and build up the
+  definition of the key methods.
+* Validate on most used modules in the ecosystem
+* Complete performance analysis to confirm accepted level of overhead.
+
+Notes/details on other related work and the current experimental progress
+on implementing this eps can be found in:
+https://docs.google.com/document/d/1bX9SZKufbfhr7ncXL9fFev2LrljymtFZdWpwmv7mB9g/edit#heading=h.ln9aj9qjo969
+
+# API definition
+
+**WORK IN PROGRESS, DRIVEN BY MODULES PORTED SO FAR**
+
+This API will be built up from a minimal set. Based on this set
+we have been able to port the NaN example and level down.  Additions
+will be made as we port additional modules.
+
+There has been enough progress that instead of listing the API methods
+here we will point to the header files in the working PoC repos so it
+remains up to date:
+
+Core methods:
+* [node_jsvmapi_types.h](https://github.com/nodejs/abi-stable-node/blob/api-prototype-6.2.0/src/node_jsvmapi_types.h)
+* [node_jsvmapi.h](https://github.com/nodejs/abi-stable-node/blob/api-prototype-6.2.0/src/node_jsvmapi.h)
+
+AsyncWorker support methods:
+
+* [node_asyncapi_types.h](https://github.com/nodejs/abi-stable-node/blob/api-prototype-6.2.0/src/node_asyncapi_types.h)
+* [node_asyncapi.h](https://github.com/nodejs/abi-stable-node/blob/api-prototype-6.2.0/src/node_asyncapi.h)
+
+
+# C++ Wrapper API
+
+The current view is that the API exported by the Node.js binary
+must be in C in order to achieve ABI stability. This [document]
+(http://www.oracle.com/technetwork/articles/servers-storage-dev/stablecplusplusabi-333927.html)
+outlines some of the challenges with respect to C++ and
+ABI stability.
+
+However, we also believe that a C++ wrapper is possible and
+would provide ease of use. The key element is that the C++ part must all
+be inlined and only use the external functions/types exported
+by the C API.  
+
+The current set of helpers to provide ease of use are defined in:
+
+[node_api_helpers.h](https://github.com/nodejs/abi-stable-node/blob/api-prototype-6.2.0/src/node_api_helpers.h)
+
+Some of the more complex parts like AsyncWorker will quite likely end up outside
+the scope of the ABI stable API itself, and be in a separate component. 
+For example, they may end up part of a version of NaN that supports the new API.
+


mhdawson:
FYI @nodejs/api, @nodejs/addon-api

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,179 @@
...
+
+## core functions
+
+### C setting javascript values
+<PRE>
+node::ni::SetMethod(ni_env, recv, name, callback)
+node::ni::SetPrototypeMethod(ni_env, recv, name, callback)
+node::ni::SetTemplate(ni_env,...)
+node::ni::SetPrototypeTemplate(ni_env,...)
+node::ni::SetInstanceTemplate(ni_env,...)
```
kkoopa: Most of these are fluff. Two functions should suffice: a `Set(primitive_or_template)` for templates and a `Set(value)` for objects.

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,179 @@
...
+
+### C getting javascript values
+<PRE>
+node::ni::ObjectHandle node:ni::getGlobalObject(ni_env)
+node::ni::ObjectHandle::getProperty(ni_env,recv, ...)
+</PRE>
+
+### C calling javascript
+<PRE>
+node::ni::callMethod(ni_env,recv,,,)
```
kkoopa: `v8::Function::Call` does not kick the next tick of the event loop, `node::MakeCallback` does (plus domain stuff). Sometimes you want the next tick, sometimes you do not. There is currently no way of just kicking the next tick from `C++`.

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,179 @@
...
+</PRE>
+
+New() takes a C type and returns a an opaque handle to the corresponding javascript
+object which is then passed into the API calls. Table will show the
+correspondance between the C types and the different types of opaque handles.
+
+| In  (C/C++)                            | out                                         |
+|----------------------------------------|---------------------------------------------|
+| char*                                  | PrimitiveStringHandle                       |
+| utf8*                                  | PrimitiveStringHandle                       |
```
kkoopa: There is no such type in `C`, is there? You cannot differentiate between utf-8 one-byte encodings by type. String handling will be a major headache whatever is decided upon and it is extremely important to get it right. Conversion between different string encondings will require an extensive library of functions and should be made as painless as possible.

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,179 @@
...
+
+## object lifecycle
+
+<PRE>
+node::ni::PushScope(ni_env,...)
+node::ni::PopScope(ni_env,...)
+node::ni::CreatePersistent(ni_env,...)
+node::ni::FreePersistent(ni_env,...)
+node::ni::CreateGlobal(ni_env,...)
+node::ni::FreeGlobal(ni_env,...)
```
kkoopa: Persistents and memory management will be very tricky in C. Also misses the weak persistents. Could consider the worthyness of uniques as well.

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,179 @@
...
+</PRE>
+
+## exception handling
+
+<PRE>
+node::ni::SetException(ni_env,...)
+node::ni: GetException(ni_env,...)
+node::ni::ClearException(ni_env,...)
```
kkoopa: I don't quite understand how this is intended to work or what the need is for `ClearException`. Why not have an exception cleared when it is gotten? What does `GetException` do?

kkoopa:
Regarding what addons to test on, don't just go for the most popular ones, pick the most complex addons, in regard to the binding part, not whatever libraries they provide. NAN's test suite, while far from comprehensive, is a good start.

There is also need to decide on what to do with Isolates. The main reason that NAN does not currently do Isolates is because of 0.10-compatibility. Should Node ever be multi-isolate, from a practical point of view? I can only think of one or two addons I know of that actually use multiple Isolates.

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,179 @@
...
+
+Notes/details on other related work and the current experimental progress
+on implementing this eps can be found in:
+https://docs.google.com/document/d/1bX9SZKufbfhr7ncXL9fFev2LrljymtFZdWpwmv7mB9g/edit#heading=h.ln9aj9qjo969
+
+# API definition
+
+**WORK IN PROGRESS, NOT ANYWHERE NEAR COMPLETE.** 
+
+This API will be built up from a minimal set as we explore the full set of Nan functions.  To start it will only include the minimal set that we need to get started.  What this means is we have not define the full API yet but instead want to give a feel of the general shape (which is not really quite there yet).
```
jasnell: s/define/defined

Fishrock123:
first reactions: doesn't seem quite as minimal as discussed?

Also, `ni` is an obscure and pretty uninformative namespace. :/ even `api` might be better?

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,179 @@
...
+
+Unfortunately in the existing implementation the V8
+(the JavaScript runtime for Node.js) APIs are exposed to modules. In 
+addition Google changes these APIs on a regular basis. This results in 
+the following issues:
+
+* Native modules must be recompiled for each version of Node.js
+* The code within modules must often be modified for a new version of
+  Node.js
+* It.s not clear which parts of the V8 API the Node.js community believes
```
bnoordhuis: It's

00X-ABI-Stable-Module-API.md
bnoordhuis: `Set(primitive_or_template)` is probably still too V8-centric to work with e.g. the SM API for defining classes.

00X-ABI-Stable-Module-API.md
bnoordhuis: I'm not sure if weakly persistent references are something that can be supported across engines.  You can implement them in SM (although it's pretty involved) but JSC's public API only has support for strongly persistent references, AFAIK.

What are the differences between Global and Persistent here?

kkoopa: Weak persistents (Finalizers) are quite necessary. node::ObjectWrap depends on them, as does node::Buffer, as do many addons, regardless of ObjectWrap or Buffers. If other engines cannot support them, how could they possibly support Node itself?

In regard to V8, Globals are movable, while Persistents are not. Globals are automatically reset when they are destroyed, Persistents are not (with the current default traits). But I do not know what the intent is regarding this API.

00X-ABI-Stable-Module-API.md
kkoopa: Not exposing templates would kill performance, would it not? I am not familiar with how SM does it, but from a pragmatic approach, should V8 not remain a first-class citizen?

jasnell:
@Fishrock123 

> first reactions: doesn't seem quite as minimal as discussed?

How so? Seems to be exactly what was discussed.

Fishrock123:
> How so? Seems to be exactly what was discussed.

Perhaps it wasn't communicated fully?

I was the under the impression we'd do only:
- call into JS & back
- call into c++ & back
- convert to basic types

As such, things like function templates _seem_ a little beyond that? Perhaps I am wrong.

Fishrock123:
Additionally I don't think the EP process along will be enough to suffice for this. (i.e. we shouldn't just write up this conceptually and sign it off)

We should aim to not expand the EP from here directly but rather get these working in a shim, make sure that suits modules, get the feedback from that, and iterate. We should make sure to make it without any guarantees until we pass a finalized version in core though since versions of it are unlikely to be maintainable for long and the API may change.

I'm totally down to help where possible, I care about this an awful lot.

00X-ABI-Stable-Module-API.md
bnoordhuis: To be clear, I'm not saying the V8 implementation shouldn't use templates, but following the V8 API too closely might make it awkward to implement it efficiently for other engines.

00X-ABI-Stable-Module-API.md
Fishrock123: Additionally, is there any reason this isn't just `ni::Call(ni_env,recv,,,)`?

kkoopa:
I'll also note that the best way of actually getting this new API into use would be to enable NAN (or a replacement) to add another (transitional) shim on top of this, ensuring backwards compatibility at least down to 0.12 or 4.0, which presumably will not have access to it. No addon maintainer wants to maintain several versions of the same code base.

jasnell:
+1. But we'd have to call it naan then ;)

kkoopa:
One more thing, make it possible to seamlessly integrate this new (stable) API with the (unstable) VM-specific API. There is no realistic way of foreseeing every possible use case or developing a practical solution that will fit everything. What I consider one of the strong aspects of NAN is that the developer is free to pick and mix between raw V8 and NAN. Some things are not achievable in any other way, but requiring a couple of specific things here and there should not condemn the rest of the code to constant breaking. For example, promises are (currently) not handled by NAN, since they are not available in Node 0.10 and there is not realistic way of backporting all of that. However, a developer is still able to easily use promises where available, while the rest of the code can (safely) use NAN. C# has a similar concept in that mixing safe and unsafe code is quite doable.

Fishrock123:
> One more thing, make it possible to seamlessly integrate this new (stable) API with the (unstable) VM-specific API. There is no realistic way of foreseeing every possible use case or developing a practical solution that will fit everything. 

I think the goal will probably end up making sure at very least node core can run ontop of it. (Since that's part of VM neutrality anyways.)

jeisinger:
I'm a bit surprised by the direction this took. I was under the impression that we agreed on a different plan in the F2F meeting, namely to take the NAN module and replace the remaining v8 dependencies with new wrappers, i.e., v8::Localv8::Object should become something like nan::Localnan::Object.

The direction this is now taking seems to be a newly designed C-style API.

Deviating that much from NAN will make it unnecessarily difficult for native modules to transition over. A C-style API precludes any inlining based optimizations and puts additional cognitive burden on module authors that will have to explicitly manage lifetime of handles as opposed to relying on the Local dtor deregistering them etc..

Another thing we agreed upon was that this API should insulate native modules from the VM - using the API inside node as a means of becoming fully VM neutral was an explicit non-goal, as the latter will require fundamentally different APIs.

jasnell:
+1 that the primary goal was to have an API that protects module developers
from the internal VM changes. It's absolutely correct that providing a new
internal abstraction was a non-goal for this initially -- if it happens to
be usable in that way, then awesome, but if not, the focus needs to be on
making sure module developers are protected.

As far as the relationship with nan is concerned, the idea was certainly to
start with nan and iterate incrementally to begin abstracting the parts
where V8's API bleeds through.

00X-ABI-Stable-Module-API.md
mhdawson: GetException would return the exception which has already been set and is pending.  ClearException is separate because you might want to be able to know there was an exception (with GetException) and what it was, but instead of clearing it, bail out from the C side and let it be handled as part of the transition back to javascript.

00X-ABI-Stable-Module-API.md
mhdawson: We  should have left out FreeGlobal() as I think the distinction is that a Persistent lives until you explicitly say it can be freed, while a Global lives forever.  We could just have the Persistent variant but having both does allow the module to give some additional information that the engine can use to optimize

00X-ABI-Stable-Module-API.md
kkoopa: Could that not be done by setting the exception again? or would that misrepresent the stack trace?

00X-ABI-Stable-Module-API.md
mhdawson: My fault here, was trying to show a few examples and picked utf8 as that is one from v8, but obviously did not get it right here.

00X-ABI-Stable-Module-API.md
mhdawson: thought callMethod was clearer, but no objection to using Call instead.

00X-ABI-Stable-Module-API.md
mhdawson: In terms of the boundary between the methods exported by the node.js binary and how the module calls them, we need a strong separation that I believe will limit the use of templating at that level.  We can't have engine code being incorporated into the module and still achieve ABI stability.

In terms of the api itself, however, the use of templates should still be able to be used to make the API easier to use but the code pulled in from the templating would be limited to code that is part of the API as opposed to the engine code.  

There are performance implications so part of the work will be validating that for most modules the level of performance achieved is acceptable.

mhdawson:
@jeisinger 

1) When we looked at the functions in Nan, they were along these lines, while the unwrapped parts of v8 were more pure C++.  So in extending the Nan methods this way seemed natural.  I think the first deviation in what is there that was not already that way in Nan is:

   node::ni::callMethod(ni_env,recv,,,)

as these others like: Nan::New() were already there.

The other areas would be object life cyle and exception handling, but would those be more than defining objects that the same methods could be called on ?  For pushScope/popScope for lifecyle the answer is probably yes, the others I'm not so sure.

In terms of using templates etc, I think that is still possible in either case, and the in-lining needs to stop before any engine code gets pulled into the module.  

Agreed, that this looks more C like than when we discussed at the meeting but I think that came out of the existing methods we pulled over from Nan.  We can iterate to move in the right direction.

2) Agree that  "using the API inside node as a means of becoming fully VM neutral was an explicit non-goal for now"

@kkoopa agreed that a Nan shim to allow this to be used transparently is one of the overall goals

@Fishrock123 agreed we can't just write this up and sign off.  We just want to give some visibility along this way.  This eps should stay in draft while we iterate and be update frequently

jeisinger:
It might be helpful to take an existing module and convert it over to the proposed API (at least parts of it to demonstrate how the new API is supposed to be used, and to get a better feeling about the changes in ergonomics this would mean).

00X-ABI-Stable-Module-API.md
kkoopa: Now I am even more confused. We were discussing V8's [templates](https://developers.google.com/v8/embed#templates) which are used to build functions and other objects. They are not a matter of ease-of-use, but a necessity for building functions and other objects with V8. Somehow I get the impression that you are thinking of C++ templates, which are an entirely different matter, having no connection to V8's templates.

mhdawson: Right I was thinking of C++ templates, v8's templates need to be completely separate from the API as we cannot pull in any engine specific code into the modules.  In the implementation for the API provided and exported by the node.js binary,  the implementation of the API can use v8 specifics but none of that can leak out to the API itself.

Fishrock123: Doesn't everyone seem to hate C++ templates or something?

00X-ABI-Stable-Module-API.md
mhdawson: It might be possible to get then reset but I'm thinking the more normal case would be to check if there is an exception and then simply return to allow it to be handled.  Another alternative would be to add a CheckException() and then make GetException both get and clear, while CheckException just returns it if one is pending.  You could then get rid of ClearException

00X-ABI-Stable-Module-API.md
mhdawson: @Fishrock123 I think its a love it or hate it kind of thing. There are some people who argue strongly that using them and C++ will make the API easier to user.  On the other side some people find the C style API easier to deal with.

mhdawson:
@jeisinger that is what I think the next step is as well. Write the code for a minimal set of the APIs and then show it working with a simple module to make things more concrete.

mhdawson:
I'll let this settle for a day or more and then incorporate the "simple" updates (typos, clarifications, etc.)

00X-ABI-Stable-Module-API.md
Fishrock123: imo, go with what is easiest to understand coming from JavaScript while remaining semantically good in most use-cases.

ianwjhalliday:
@kkoopa Which addons are those?  Sounds like they are good to have on hand for investigation once we get the balling rolling on working prototype.

> There is also need to decide on what to do with Isolates. The main reason that NAN does not currently do Isolates is because of 0.10-compatibility. Should Node ever be multi-isolate, from a practical point of view? I can only think of one or two addons I know of that actually use multiple Isolates.

kkoopa:
webworker-threads

Fishrock123:
> Should Node ever be multi-isolate, from a practical point of view

Yes, so that workers are a thing we can do since we've been wanting to for a while but it has been too complex to review effectively. See: https://github.com/nodejs/node/pull/2133

00X-ABI-Stable-Module-API.md
trevnorris: I'm with @kkoopa on this. If an engine doesn't support weak persistents then how are they going to implement node? Just search the code for `MakeWeak` and you'll see it's used quite a lot.

README.md
```diff
@@ -58,15 +58,25 @@ All EP documents must be MIT licensed.
 ## Progress of an EP
 
 All EPs will be committed to the repository regardless of their acceptance.
-The initial status shall be **"DRAFT"**.
+The initial status shall either be **"INCUBATING"** or **"DRAFT"**.
+
+If the status is **"INCUBATING"** that means that it is expected that the
```
trevnorris: Why is this necessary? The incubation period is the time spent in the PR. If the EP is still so volatile that it expects to have changes or additions over time then it shouldn't land.

If this is the initial api and later there's an addition then the addition should go in a new EP. These aren't design documents that are meant to reflect a feature's current state.

trevnorris:
Please replace `<PRE>` with three grave symbols. (Backticks)

00X-ABI-Stable-Module-API.md
bnoordhuis: Let me amend my previous statement by saying that you can create objects with finalizers in JSC (which I think is the primary use case for weakly persistent references) but the V8 semantics of taking an arbitrary JS value and enabling/disabling weakness on the fly is not something that it efficiently supports (or, I suspect, at all.)

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,179 @@
...
+used as an ordinary Node.js module (for more details see
+https://nodejs.org/api/addons.html)
+
+Unfortunately in the existing implementation the V8
+(the JavaScript runtime for Node.js) APIs are exposed to modules. In 
+addition Google changes these APIs on a regular basis. This results in 
+the following issues:
+
+* Native modules must be recompiled for each version of Node.js
+* The code within modules must often be modified for a new version of
```
ofrobots: 'must often' is too strong. Node.js 6.x didn't have API changes and didn't require module code to be modified. I would suggest wording this as 'The code within modules may need to be modified for a new version of Node.js'.

README.md
mhdawson: @trevnorris this mistakenly got pushed as part of an update can you comment here instead:  https://github.com/nodejs/node-eps/pull/22

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,179 @@
...
+  Node.js
+* It.s not clear which parts of the V8 API the Node.js community believes
+  are safe/unsafe to use in terms of long term support for that use.
+* Modules written against the V8 APIs may or may not be able to work
+  with alternate JavaScript engines when/if Node.js supports them.
+
+The Nan project (https://github.com/nodejs/nan) can be used to limit
```
nebrius: nit: s/Nan/NAN. Maybe also add the full words in the acronym, e.g. `The Native Abstractions for Node (NAN) project`?

00X-ABI-Stable-Module-API.md
```diff
@@ -80,222 +80,290 @@ https://docs.google.com/document/d/1bX9SZKufbfhr7ncXL9fFev2LrljymtFZdWpwmv7mB9g/
...
-</PRE>
-
+**WORK IN PROGRESS, DRIVEN BY MODULES PORTED SO FAR**
+
+This API will be built up from a minimal set. Based on this set
+we have been able to port the NaN example and level down.  Additions
+will be made as we port additional modules.
+
+### Types
+````
```
ianwjhalliday: Could specify this is C for syntax highlighting

00X-ABI-Stable-Module-API.md
```diff
@@ -80,222 +80,290 @@ https://docs.google.com/document/d/1bX9SZKufbfhr7ncXL9fFev2LrljymtFZdWpwmv7mB9g/
...
+would provide ease of use. The key element is that the C++ part must all
+be inlined and only use the external functions/types exported
+by the C API.  
+
+What follows are some examples (incomplete/immature) mirroring NaN when
+possible on what some of the C++ wrapper might include. Its not as far along
+as the C API as we've only added things that were needed in porting the
+modules we've worked on so far.  In the end we expect a more complete set.
+
+````
```
ianwjhalliday: and likewise C++

ianwjhalliday:
Updated document looks great

ianwjhalliday:
:+1:

mhdawson:
@nodejs/collaborators could I get a few reviews/L.G.T.M.s or comments.  Discussion in last CTC meeting was that it can land as draft status even if we have not capture the final state of the API.

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,369 @@
...
+
+**WORK IN PROGRESS, DRIVEN BY MODULES PORTED SO FAR**
+
+This API will be built up from a minimal set. Based on this set
+we have been able to port the NaN example and level down.  Additions
+will be made as we port additional modules.
+
+### Types
+```C
+typedef struct napi_env__ *napi_env;
```
saghul: Not sure if bikeshed, but I find not typedef-ing pointers a lot easier to read. If I see some `napi_env e` I may think it's a typedef struct, vs `napi env* e` which is clearer.

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,369 @@
...
+    napi_symbol,
+    napi_object,
+    napi_function,
+};
+```
+### Functions
+
+```C
+// Getters for defined singletons
+NODE_EXTERN napi_value napi_get_undefined_(napi_env e);
```
saghul: trailing underscore?

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,369 @@
...
+    napi_string,
+    napi_symbol,
+    napi_object,
+    napi_function,
+};
+```
+### Functions
+
+```C
+// Getters for defined singletons
```
saghul: how about:

```
enum napi_singleton {
    napi_undefined,
    napi_null,
    ....
}
NODE_EXTERN napi_value napi_get_singleton(napi_env e, napi_singleton type);
```

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,369 @@
...
+NODE_EXTERN napi_value napi_get_undefined_(napi_env e);
+NODE_EXTERN napi_value napi_get_null(napi_env e);
+NODE_EXTERN napi_value napi_get_false(napi_env e);
+NODE_EXTERN napi_value napi_get_true(napi_env e);
+NODE_EXTERN napi_value napi_get_global_scope(napi_env e);
+
+
+// Methods to create Primitive types/Objects
+NODE_EXTERN napi_value napi_create_object(napi_env e);
```
saghul: how about using variadic args?

```
NODE_EXTERN napi_value napi_create(napi_env e, napi_something thing, ...);
```

And depending on what `thing` is more or less parameters are required. This would reduce the API surface.

saghul:
Had a quick read and left some comments, hope that helps.

00X-ABI-Stable-Module-API.md
ianwjhalliday: Since they're opaque types the idea is the user isn't supposed to consider what they are exactly, and imo all the stars that would be typed are redundant once you get used to the types being opaque handles.  Indeed we can bikeshed this point later on when a complete API is being reviewed for PR.  For now though we are still prototyping the idea.

00X-ABI-Stable-Module-API.md
ianwjhalliday: I'm not sure reducing the number of exports necessarily helps to reduce API surface.  This would just turn the API into a different form.  Again something we should note and debate later on in API review.

00X-ABI-Stable-Module-API.md
ianwjhalliday: @mhdawson can you fix this typo?  I think this is also in the actual code right now too.

00X-ABI-Stable-Module-API.md
ianwjhalliday: Again we can discuss in API review later on.

saghul: What is it that should be reviewed at this stage then? I'm afraid I may have misunderstood something here.

00X-ABI-Stable-Module-API.md
mhdawson: Will fix

00X-ABI-Stable-Module-API.md
mhdawson: The goal at this point is to land the ep as a draft.  We expect there will be changes to the API's as we build up experience in using them to port/performance etc.  Once its in as a draft them we can have PR's that update based on the experimentation or even just suggestions.  So its not that suggestions aren't valid but that we are not trying to achieve the final shape in the DRAFT state.    Even better would be PR's to the code itself.

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,369 @@
...
+
+The module ecosystem is one of the features making Node.js the fastest
+growing runtime.  An important subset of modules are those which have a
+native component. In specific cases it is important to be able to leverage native
+code in order to be able to:
+
+* use existing third party components written in C/C++
+* access physical devices, for example a serial port
+* expose functionality from the operating system not otherwise available
+* achieve higher speed than possible in native JavaScript
```
Qard: Might also be worth noting the possibility of pulling CPU-intensive work out-of-band from the JS thread. That's sort of "speed" related, but maybe different enough that it warrants an explicit mention.

Qard:
One thing I'd like clarified is if the C++ API sits atop the C API, or vice versa.

If the C++ is a thin layer over the C, the C++ could perhaps be shipped as a header file, allowing the benefits of C++ API friendliness without dealing with the ABI stability concerns.

ianwjhalliday:
@Qard that's correct, the intention is a C++ layer that is shipped entirely in a header and is built upon the C exports only, providing no exports of its own, for ABI stability.

Agree this should be clearly stated in the document.

@mhdawson perhaps we should mention this as a bullet in the `Proposed Action` section?  We can also be explicit about this intent in the `C++ Wrapper API` section.

Qard:
Great, thanks for the clarification! :smiley_cat:

mhdawson:
Ok, plan to address the most recent comments and the hopefully get enough oks to land.  Once landed additional changes can then be more easily addressed through pull requests.

00X-ABI-Stable-Module-API.md
mhdawson: Noted, will look at  using the variadic args as well evolve the POC but will leave doc as is for now

mhdawson:
I think this section in the doc:

```
However, we also believe that a C++ wrapper is possible and would provide ease of use. The key element is that the C++ part must all be inlined and only use the external functions/types exported by the C API. 
```

covered the question from @qard so don't think we need to update to the C++ Wrapper API section.

Added this to the Proposed Action section:
- The API provided by the Node.js runtime will be in C. In
  addition a C++ wrapper which only contains inline funcctions
  and uses the C APIs so that the changes in the C++ wrapper
  would not require recompilation.

@Quad @ianwjhalliday @saghul @trevnorris  @jasnell can you take another look and give me an LGTM if appropriate to land as "DRAFT" knowing it will change was we work through the PoC.

saghul:
LGTM as a draft.  One thing that the document should contain IMHO is the ABI guarantees. As in, when is the ABI allowed to be broken / modified? I guess with a semver major, right?

mhdawson:
Ideally we don't break ABI at all.  Semver major would be absolute minimum but we'd want it to be rare enough that semver major is one of the weaker requirements.

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,373 @@
...
+    virtual void HandleErrorCallback()
+    void SetErrorMessage(const char *msg) 
+
+    const char* ErrorMessage() 
+  };
+
+  inline void AsyncExecute(uv_work_t* req)
+  inline void AsyncExecuteComplete(uv_work_t* req)
+  inline void AsyncQueueWorker(AsyncWorker* worker) 
+```
```
mcollina: I am a bit conflicted about `AsyncWorker`. We are currently making it "part of core", when it is currently not.
However it is highly used in C++ addons, so I think it will be highly useful.

Can we keep all of these in another module for the time being? It's not bad to provide those via a dependency that we can version independently.

mhdawson: Agreed that we should manage the additions beyond the C API carefully and that in this case they may belong elsewhere.  I'll add a comment indicating that they will likely end up separate from the API itself.

mhdawson:
@trevnorris any chance I can get a second L.G.T.M from you to land as a draft ?

mhdawson:
@nodejs/api can I get review/approval from anybody else in the group ?

agnat:
Before diving into technical details some thoughts about the document. It has been edited many times without reworking the text in full. Here are some suggestions to make your work more accessible. *borrows a TC members hat*

The thing I'm missing most is information about the design decision process.  I would like to better understand the "Why?"s and "How?"s _behind_ the design. For example, I've seen a brief discussion of alternatives but found no comparison of their technical merits which would explain why this approach was taken. I'd like to know what options have been explored. This proposal contains important design decisions that will affect all node addons. I would like the proposal to mention not only the advantages but also the disadvantages of the approach taken.

I also recommend to rework the "Proposed Action" and "Background" sections. "Proposed Action" lists goals and "Background" contains a roadmap aka. "Proposed Actions" . Maybe be like:

#### Goals:

Describe what we would like to achieve in an abstract way, expanding each bullet point with a sentence or two. For example it is not clear to me what "VM agnostic" entails. First it sounds like different versions of v8, later (and digging deeper) it seems to include different JavaScript engines like SpiderMonkey &c. I definitely see evidence of feature creep and the actual scope is not entirely clear to me.

I also feel that "ABI stability" is (way) to vague. It's such a complex subject. I'd comment on what sort of independence we want to achieve. I'd also add a sentence about what ABI related issues will remain. Let's make it more thorough. The [oracle document](http://www.oracle.com/technetwork/articles/servers-storage-dev/stablecplusplusabi-333927.html) helps to understand the issue, but this proposal should draw the line.

In the introduction there is a list of undesirable features of the status quo. Which of those are we going to address?

Oh, and list the non-goals too. Any _future_ goals that this proposal takes into account?

#### Proposed Action

Describe how we achieve the goals above. Reassemble and expand the roadmap. Maybe add a link to the nodejs/abi-stable-node#2 (and other?) milestone ticket.

#### Background:

Try to get rid of this section. Lift the important pieces, those that drive this proposal from the documents linked and include them here consolidating your work. In particular the [google doc](https://docs.google.com/document/d/1bX9SZKufbfhr7ncXL9fFev2LrljymtFZdWpwmv7mB9g/edit#heading=h.ln9aj9qjo969) contains a lot of information that I feel belongs into the proposal. AFAICT, the rest is mostly "Goals" or "Proposed Action" stuff.

Sorry to propose all these text changes, but I think it is an ambitious proposal and it deserves more clarity.

agnat:
The proposal describes the design by presenting two interfaces: 1. The C API and 2. the user (or addon developers) interface. I think a better view is to describe it in terms of subsystems or layers. 
1. The abstraction layer between v8/node and the C API
2. A header only user interface layer with the C API on one side and a v8/node/nan style API on the user end.

It is these subsystems we should talk about, not the interfaces. Note that in this view the actual shape of the C API is largely irrelevant. It is internal. I think there is a comment somewhere along the lines: "Let's just let the C API evovle as needed. Later optimize it to be very generic and universal, so we don't have to break it again." I'm paraphrasing, but I believe this is the right approach. So, let's forget the actual interface for a second.

Also, forget the details of the other interface, the C++ wrapper API. We already know what it should look like. In a first step it should be as close as possible to what we currently have: v8 style type system, node and nan utilities, similar syntax, similar features. Just state your goals. No point in presenting an incomplete interface.

What _really_ matters is the code in these new layers. That is our investment. We commit to develop and maintain these new subsystems. An honest attempt should be made to outline the amount and the complexity of the code, because that is one thing the TC needs to know.

Instead of presenting the interfaces I suggest linking to (PRs on) the current PoC implementations, so we can take a look:
- [nodejs/abi-stable-node](https://github.com/nodejs/abi-stable-node)
- [nodejs/abi-stable-node-addon-examples](https://github.com/nodejs/abi-stable-node-addon-examples)
- can not find the level down PoC branch

mhdawson:
@agnat the kinds of suggestions you are making is exactly why I'd like to get this landed as a DRAFT. You could then submit PR's to make the enhancements you are suggesting and they could be discussed in the PR.

jasnell:
This LGTM, we can update with follow on PRs

mhdawson:
@agnat I'm going to go ahead and land the current version and then look to address your comments through PRs. Do you want to submit PRs for the suggested changes or would you like me to do that based on your comments ?

agnat:
@mhdawson that is not how it is supposed to work. People spent time and effort working through this asking very valid questions in the comments above. Not only me.

Answering these questions in the comment section is _not enough_. Postponing them till later is not ok. These questions arise while reading the proposal. The _proposal_ should answer them. Please respect the work _reviewers_ put into this. Include their feedback and improve the document.

I've been following this PR since [pinged on day one](https://github.com/nodejs/node-eps/pull/20#issuecomment-215889878). I postponed the review time and again _because_ the document is no help. Diving in anyway 17some days ago it took me more than 16 hours of work to develop a mental picture of what _really_ is about to happen. It is the documents job to describe the system. Instead I had to work it out myself.

I'm ready to help with the document down the road, but as it stands it simply lacks the structure to do that.
1. The interface centric view will _never_ be final until the thing is done. Describe the (path to a) system instead.
2. Stating that `$X` is incomplete in the document guarantees major misunderstandings. Nobody knows if you skipped a feature because you decided not to support it or if it is an oversight. Describe the scope.
3. Mention performance and maintenance implications.
4. Answer the questions above in the document.

Please provide a proper technical proposal we can work on. One we can actually discuss because it has some meat.

mhdawson:
@agnat, I am going to need more insight into what you are suggesting before jumping into the work you are suggesting.

I think a higher bandwidth discussion between us would help me understand where/what you'd like changed.  I'm thinking a 30 minute discussion were we go through and come up with the main headings in the doc together would go a long way in helping me understand the concrete changes that will address you concerns.    Would that be possible ? 

We can do that either one on one or during the weekly get together held by the group working on the PoC (Fridays at 2EST).

brody2consult:
Why define the C++ interface in the same place? I would rather see the C++ wrapper defined and/or implemented somewhere else. Simpler parts working together (UNIX philosophy). As an idea: could we make a new NAN version that provides the C++ wrapper?

brody2consult:
I would also like to see a couple more API functions/enhancements:
1. extern "C" equivalent to `int node::Start(int argc, char *argv[])`
2. API function to start node instance in a non-default uv_loop (with a new V8 Isolate)

1 is needed by C embedders while 2 could help enable true multithreading on Node.js.

I experimented with I would be happy to submit a new issue and maybe a PR on https://github.com/nodejs/abi-stable-node to explain what I mean by 2.

+1 for getting something landed, maybe something simple so people can start submitting PRs. What is the status? Any conclusions between @mhdawson / @agnat?

_P.S. I would be happy to raise both 1 and 2 as new issue(s). :-1: for not landing a DRAFT._

Fishrock123:
Tagging with `ctc-agenda` as discussed in the Node Interactive NA VM Summit meeting.

jasnell:
@nodejs/ctc : This EP was discussed extensively during the VM Summit today at the Node Interactive Collaborator Summit. While this is still very much a work in progress, there has been sufficient progress towards a working prototype and a well defined direction. We should make a decision about whether this EP should be accepted as a draft (and this PR landed). The recommendation of the VM Summit participants is that this should be accepted as a draft.

Trott:
@nodejs/ctc Note that the CTC is being asked to review this EP, so please review it carefully if you are able.

@jasnell @Fishrock123 Any reason not to post an issue in the CTC repo and call for a vote right now? Or is there a reason the process needs to wait for the actual meeting?

jasnell:
@trott I'd like to make sure that there's at least an opportunity for the CTC members to ask any clarification questions but opening a vote is a good idea.

00X-ABI-Stable-Module-API.md
bnoordhuis: Using typedef'd pointers has the issue that they will silently decay to `void*` when used in a `void*` context.  Can be avoided by wrapping the pointer in a struct, like so:
```c
typedef struct napi_value { struct napi_value *v; } napi_value;
```

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,377 @@
...
+typedef struct napi_escapable_handle_scope__ *napi_escapable_handle_scope;
+typedef struct napi_propertyname__ *napi_propertyname;
+typedef const struct napi_func_cb_info__ *napi_func_cb_info;
+typedef void (*napi_callback)(napi_env, napi_func_cb_info);
+typedef void napi_destruct(void* v);
+
+
+struct napi_method_descriptor {
+  napi_callback callback;
+  char* utf8name;
```
bnoordhuis: `const char*`?

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,377 @@
...
+NODE_EXTERN napi_value napi_create_array_with_length(napi_env e, int length);
+NODE_EXTERN napi_value napi_create_number(napi_env e, double val);
+NODE_EXTERN napi_value napi_create_string(napi_env e, const char*);
+NODE_EXTERN napi_value napi_create_string_with_length(napi_env e, const char*,
+                                                               size_t length);
+NODE_EXTERN napi_value napi_create_boolean(napi_env e, bool b);
+NODE_EXTERN napi_value napi_create_symbol(napi_env e, const char* s = NULL);
+NODE_EXTERN napi_value napi_create_function(napi_env e, napi_callback cbinfo);
+NODE_EXTERN napi_value napi_create_error(napi_env e, napi_value msg);
+NODE_EXTERN napi_value napi_create_type_error(napi_env e, napi_value msg);
```
bnoordhuis: How will these functions signal error?  E.g. what if `napi_create_object()` can't create an object because the VM is out of memory or in a state where it can't be entered?

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,377 @@
...
+                                           char* buf, const int buf_size);
+NODE_EXTERN int32_t napi_get_value_int32(napi_env e, napi_value v);
+NODE_EXTERN uint32_t napi_get_value_uint32(napi_env e, napi_value v);
+NODE_EXTERN int64_t napi_get_value_int64(napi_env e, napi_value v);
+NODE_EXTERN bool napi_get_value_bool(napi_env e, napi_value v);
+
+NODE_EXTERN int napi_get_string_length(napi_env e, napi_value v);
+NODE_EXTERN int napi_get_string_utf8_length(napi_env e, napi_value v);
+NODE_EXTERN int napi_get_string_utf8(napi_env e, napi_value v,
+                                     char* buf, int bufsize);
```
bnoordhuis: Why int instead of size_t?

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,377 @@
...
+NODE_EXTERN bool napi_is_array(napi_env e, napi_value v);
+NODE_EXTERN uint32_t napi_get_array_length(napi_env e, napi_value v);
+
+// Methods to compare values
+NODE_EXTERN bool napi_strict_equals(napi_env e, napi_value lhs, napi_value rhs);
+
+
+// Methods to work with Functions
+NODE_EXTERN void napi_set_function_name(napi_env e, napi_value func,
+                                        napi_propertyname napi_value);
```
bnoordhuis: How would this work with an engine like spidermonkey where you would normally declare the name upfront?

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,377 @@
...
+                                          int argc, napi_value* argv);
+
+
+// Methods to work with napi_callbacks
+NODE_EXTERN int napi_get_cb_args_length(napi_env e, napi_func_cb_info cbinfo);
+NODE_EXTERN void napi_get_cb_args(napi_env e, napi_func_cb_info cbinfo,
+                                  napi_value* buffer, size_t bufferlength);
+NODE_EXTERN napi_value napi_get_cb_this(napi_env e, napi_func_cb_info cbinfo);
+// V8 concept; see note in .cc file
+NODE_EXTERN napi_value napi_get_cb_holder(napi_env e, napi_func_cb_info cbinfo);
```
bnoordhuis: Do other engines distinguish between this and holder? (EDIT: Or is that what "V8 concept" refers to?)

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,377 @@
...
+NODE_EXTERN bool napi_is_construct_call(napi_env e, napi_func_cb_info cbinfo);
+NODE_EXTERN void napi_set_return_value(napi_env e, napi_func_cb_info cbinfo,
+                                       napi_value v);
+
+
+// Methods to support ObjectWrap
+NODE_EXTERN napi_value napi_create_constructor_for_wrap(napi_env e,
+                                                        napi_callback cb);
+NODE_EXTERN void napi_wrap(napi_env e, napi_value jsObject, void* nativeObj,
```
bnoordhuis: Will this work in all engines?

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,377 @@
...
+                                             napi_escapable_handle_scope s);
+NODE_EXTERN napi_value napi_escape_handle(napi_env e,
+                                          napi_escapable_handle_scope s,
+                                          napi_value v);
+
+
+// Methods to support error handling
+NODE_EXTERN void napi_throw(napi_env e, napi_value error);
+NODE_EXTERN void napi_throw_error(napi_env e, char* msg);
+NODE_EXTERN void napi_throw_type_error(napi_env e, char* msg);
```
bnoordhuis: `const char*`

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,377 @@
...
+
+// These are particularly rough, needs more analysis
+typedef void (*napi_try_callback)(napi_env e, void* data);
+typedef void (*napi_catch_callback)(napi_env e, void* data);
+NODE_EXTERN bool napi_try_catch(napi_env e, napi_try_callback t,
+                                napi_catch_callback c, void* data);
```
bnoordhuis: I think most JS engines have an 'is exception pending?' function; V8 is the outlier with its TryCatch construct.  API strawman:
```c
// Wrapper around a v8::TryCatch.  Dummy with other engines.
typedef struct napi_trycatch { struct napi_trycatch *v; } napi_trycatch;
NODE_EXTERN void napi_trycatch_new(napi_env e, napi_trycatch *trycatch);
// Returns exception object or undefined.
NODE_EXTERN napi_value napi_trycatch_exception(napi_env e, napi_trycatch trycatch);
NODE_EXTERN void napi_trycatch_delete(napi_env e, napi_trycatch trycatch);
```

00X-ABI-Stable-Module-API.md
```diff
@@ -0,0 +1,377 @@
...
+The current view is that the API exported by the Node.js binary
+must be in C in order to achieve ABI stability. This [document]
+(http://www.oracle.com/technetwork/articles/servers-storage-dev/stablecplusplusabi-333927.html)
+outlines some of the challenges with respect to C++ and
+ABI stability.
+
+However, we also believe that a C++ wrapper is possible and
+would provide ease of use. The key element is that the C++ part must all
+be inlined and only use the external functions/types exported
+by the C API.
```
bnoordhuis: The problem with inlining is that there is no way to release bug fixes without everyone needing to recompile.

00X-ABI-Stable-Module-API.md
mhdawson: Adding error handling is the next main milestone that the time is working on.

00X-ABI-Stable-Module-API.md
mhdawson: So far we've got it working for v8 and Chakra.

00X-ABI-Stable-Module-API.md
mhdawson: That is true, the sugar becomes part of the module.

00X-ABI-Stable-Module-API.md
mhdawson: Yes, the note is saying its a v8 concept.

00X-ABI-Stable-Module-API.md
mhdawson: opened https://github.com/nodejs/abi-stable-node/issues/25 to update current implementation.

00X-ABI-Stable-Module-API.md
mhdawson: opened https://github.com/nodejs/abi-stable-node/issues/25 to update current implementation.

00X-ABI-Stable-Module-API.md
mhdawson: opened https://github.com/nodejs/abi-stable-node/issues/25 to update current implementation.

00X-ABI-Stable-Module-API.md
mhdawson: so still a work in progress to see if it makes sense across engines, so far ok for v8 and Chakra.

00X-ABI-Stable-Module-API.md
mhdawson: Created to https://github.com/nodejs/abi-stable-node/issues/26 to investigate/figure out  if we need a different API here.

00X-ABI-Stable-Module-API.md
mhdawson: Create https://github.com/nodejs/abi-stable-node/issues/27 to discuss as part of prototype

mhdawson:
@bnoordhuis  has a number of good suggestions.  I also know that the API will still change as the PoC continues so I opened issues in the abi-stable-node repo to address those.  

Earlier, before we had progressed the PoC to working code it was useful to have the functions in the eps to give people a "feel" for what it might look like.  At this point I'm wondering if we should remove it and replace it with a pointer to the header in the abi-stable-node repo: https://github.com/nodejs/abi-stable-node/blob/api-prototype-6.2.0/src/node_jsvmapi.h as it will continue to evolve.

jasnell:
+1 to landing this EPS as a draft and iterating on from there

mhdawson:
Discussed in CTC meeting last week.  Agreed to land unless objects posted here.  Going to land now.

mhdawson:
Squashed commits.

mhdawson:
Landed as 735b7765a4a2ed893cafc7afa0426c912ed1d09e

trevnorris:
Just noticed that it wasn't added to the index. Added it for you in https://github.com/nodejs/node-eps/commit/54b199c4eb58ff8b3ac8f273bd52af727bfc9dac

PR #22, opened 2016-05-03
doc: Update README.md to include WIP status

mhdawson:
Based on the EP for the stable module ABI I think
an additional status would be useful to call out
those which we believe are ready for final review
versus those that we expect to iterate on before
that final review should take place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,17 @@ All EP documents must be MIT licensed.
 ## Progress of an EP
 
 All EPs will be committed to the repository regardless of their acceptance.
-The initial status shall be **"DRAFT"**.
+The initial status shall either be **"INCUBATING"** or **"DRAFT"**.
+
+If the status is **"INCUBATING"** that means that it is expected that the
+EP will be iterated on over a period of time before it becomes a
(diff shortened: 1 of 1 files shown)


jasnell:
I've always disliked "incubating" as a status. Perhaps just "pre-draft"?

mhdawson:
or maybe "WIP" for work in progress ? otherwise "pre-draft" is fine with me.

evanlucas:
I like WIP

jasnell:
Work in progress works for me too

mhdawson:
Updated to use WIP

mhdawson:
Any additional comments ?

@trevnorris I know you had some on the other PR where the change had been accidentally included.

trevnorris:
I'm not sure if I completely agree with how this works. Documents can land as DRAFT while they're being revised. A WIP to me means that the PR is asking for help to review/fix/etc. That's easily enough placed in the title of the PR. I don't like the idea of a document landing with WIP, since it denotes that it's definitely not something even potentially accepted. A document should land unless it's been agreed upon that it could be implemented.

mhdawson:
Long since stale.  node-eps has been sunset.

PR #28, opened 2016-06-01
proposal: WHATWG URL standard implementation

jasnell:
The [WHATWG URL Standard](https://url.spec.whatwg.org) specifies updated syntax,
parsing and serialization of URLs as currently implemented by the main Web Browsers.
The existing Node.js `url` module parsing and serialization implementation currently 
does not support the URL standard and fails to pass about 160 of the standard tests.

This proposal is to implement the WHATWG URL Standard by introducing a new
`URL` class off the `url` module (e.g. `require('url').URL`).

The _existing_ `url` module would **remain** unchanged and there should be no
backwards compatibility concerns.
## Example

``` js
const url = new URL('http://user:pass@example.org:1234/p/a/t/h?xyz=abc#hash');

console.log(url.protocol;      // http:
console.log(url.username);     // user
console.log(url.password);     // password
console.log(url.host);         // example.org:1234
console.log(url.hostname);     // example.org
console.log(url.port);         // 1234
console.log(url.pathname);     // /p/a/t/h
console.log(url.search);       // ?xyz=abc
console.log(url.searchParams); // SearchParams object
console.log(url.hash);         // hash

// The SearchParams object is defined by the WhatWG spec also
url.searchParams.append('key', 'value');

console.log(url);
 // http://user:pass@example.org:1234/p/a/t/h?xyz=abc&key=value#hash
```

### DIFF
--- a/000-index.md
+++ b/000-index.md
@@ -2,3 +2,4 @@
 
 * [Public C++ Streams](001-public-stream-base.md)
 * [ES6 Module Interoperability](002-es6-modules.md)
+* [WHATWG URL API](003-url.md)
--- a/003-url.md
+++ b/003-url.md
@@ -1,0 +1,177 @@
+| Title  | Implement WHATWG URL Spec   |
+|--------|-----------------------------|
+| Author | @jasnell                    |
+| Status | DRAFT                       |
+| Date   | 2016-08-11T10:00:00-07:00   |
+
+## Description
+
+The [WHATWG URL Standard](https://url.spec.whatwg.org/) specifies updated
+syntax, parsing and serialization of URLs as currently implemented by the main
+Web Browsers. The existing Node.js `url` module parsing and serialization
+implementation currently does not support the URL standard and fails to pass
+about 160 of the WHATWG URL parsing tests.
+
+This proposal is to implement the WHATWG URL Standard by modifying the existing
+`url` module to provide an implementation of the `URL` object and associated
+APIs as defined by the WHATWG URL Parsing specification. Doing so improves the
+robustness of URL parsing, provides consistency with browser JS code, and can
+eventually allow the reduction of node.js specific APIs.
+
+Initially, the implementation would be introduced as an *undocumented*
+experimental feature exposed via a new `URL` property in the `url` module.
+
+*Note*: In Browser implementations, the `URL` constructor is a global. In the
+initial Node.js implementation, the `URL` constructor would **not** be a
+global; instead, it would be accessible via the `url` module (e.g.
+`const URL = require('url').URL`).
+
+Because the existing `require('url')` module remains otherwise unmodified,
+there should be no backwards compatibility concerns. Once the performance of
+the new URL implementation is on par with the existing `url.parse()`
+implementation, `url.parse()` will go through a normal deprecation cycle and
+the new URL implementation will become a fully documentation and supported
+feature.
+
+The current implementation can be found at: 
+
+https://github.com/nodejs/node/pull/7448
+
+## Example
+
+```js
+const URL = require('url').URL;
+const url = new URL('http://user:pass@example.org:1234/p/a/t/h?xyz=abc#hash');
+
+console.log(url.protocol);      // http:
+console.log(url.username);     // user
+console.log(url.password);     // password
+console.log(url.host);         // example.org:1234
+console.log(url.hostname);     // example.org
+console.log(url.port);         // 1234
+console.log(url.pathname);     // /p/a/t/h
+console.log(url.search);       // ?xyz=abc
+console.log(url.searchParams); // SearchParams object
+console.log(url.hash);         // hash
+
+// The SearchParams object is defined by the WHATWG spec also
+url.searchParams.append('key', 'value');
+
+console.log(url);
+ // http://user:pass@example.org:1234/p/a/t/h?xyz=abc&key=value#hash
+ 
+
+// Example using a base URL
+const url2 = new URL('/foo', url);
+console.log(url.protocol);      // http:
+console.log(url.username);     // user
+console.log(url.password);     // password
+console.log(url.host);         // example.org:1234
+console.log(url.hostname);     // example.org
+console.log(url.port);         // 1234
+console.log(url.pathname);     // /foo
+console.log(url.search);       // ''
+console.log(url.searchParams); // SearchParams object
+console.log(url.hash);         // ''
+```
+
+## APIs
+
+The public API would be as defined by the
+[WHATWG spec](https://url.spec.whatwg.org/#api).
+
+### `new URL(href, base)`
+
+The constructor implements the WHATWG basic parsing algorithm. Accessible via
+`const URL = require('url').URL`
+
+* `href` is a `string` containing the URL to parse.
+* `base` is either a `string` or a `URL` that contains the base URL to resolve
+  against while parsing.
+
+See https://url.spec.whatwg.org/#urlutils-members for detail on the properties
+of the `URL` object.
+
+#### `url.protocol`
+
+* Returns a string
+* Getter/Setter
+
+#### `url.username`
+
+* Returns a string
+* Getter/Setter
+
+#### `url.password`
+
+* Returns a string
+* Getter/Setter
+
+#### `url.host`
+
+* Returns a string
+* Getter/Setter
+
+#### `url.hostname`
+
+* Returns a string
+* Getter/Setter
+
+#### `url.port`
+
+* Returns an integer
+* Getter/Setter
+
+#### `url.pathname`
+
+* Returns a string
+* Getter/Setter
+
+#### `url.search`
+
+* Returns a string
+* Getter/Setter
+
+#### `url.searchParams`
+
+* Returns a URLSearchParams object
+* Getter
+
+#### `url.hash`
+
+* Returns a string
+* Getter/Setter
+
+#### `url.origin`
+
+* Returns a string
+* Getter
+
+#### `url.href`
+
+* Returns a string
+* Getter
+
+#### `url.toString()`
+
+* Returns a string (same as `url.href`)
+
+### `URLSearchParams`
+
+Returned by the `url.searchParams` property and provides access read and write
+access to the search params. It is defined at
+https://url.spec.whatwg.org/#interface-urlsearchparams.
+
+Accessible via `require('url').URLSearchParams`
+
+#### `searchParams.append(name, value)`
+#### `searchParams.delete(name)`
+#### `searchParams.get(name)`
+#### `searchParams.getAll(name)`
+#### `searchParams.has(name)`
+#### `searchParams.set(name, value)`
+#### `searchParams.*[Symbol.iterator]()`
+#### `searchParams.toString()`
+
+### `URL.domainToASCII(domain)`
+### `URL.domainToUnicode(domain)`


jasnell:
@nodejs/collaborators

seishun:
Perhaps we could raise interest in this by providing examples of failing tests.

jasnell:
@seishun ... https://github.com/nodejs/node/blob/master/test/known_issues/test-url-parse-conformance.js :-)

We currently fail somewhere around 140+ of the test cases in the WhatWG set.

MylesBorins:
+1

Considering that the browser are exposing the global I think it makes a whole bunch of sense. I'd be interested in helping out with this. Have you broken ground on implementation @jasnell?

Can we borrow from implementations at all?

XXX-url.md
```diff
@@ -0,0 +1,75 @@
...
+
+Because the existing `require('url')` module remains, there should be no
+backwards compatibility concerns.
+
+## Example
+
+```js
+const url = new URL('http://user:pass@example.org:1234/p/a/t/h?xyz=abc#hash');
+
+console.log(url.protocol;      // http:
```
yorkie: missing ')'

XXX-url.md
```diff
@@ -0,0 +1,75 @@
...
+| Author | @jasnell                    |
+| Status | DRAFT                       |
+| Date   | 2016-06-01T09:00:00-07:00   |
+
+## Description
+
+The WhatWG URL Standard specifies updated syntax, parsing and serialization of
+URLs as currently implemented by the main Web Browsers. The existing Node.js
+`url` module parsing and serialization implementation currently does not support
+the URL standard and fails to pass almost half of the parsing tests.
```
yorkie: @jasnell should we specify the precise number of failing cases? namely 140+ as you mentioned below :-)

jasnell:
@TheAlphaNerd ... yeah, I've got it mostly implemented already. The next step is to start running it through it's paces with tests and benchmarks and to find ways of optimizing the implementation. It's currently quite a bit slower than the existing `require('url')` module parsing.

XXX-url.md
jasnell: Yeah, I'll run the tests here shortly and update the number :-)

jasnell:
Regarding borrowing from other impls, it's entirely possible that we could borrow from chromes implementation. I'm not sure yet if theirs is a pure JS impl or not. I'll look into that.

XXX-url.md
```diff
@@ -0,0 +1,75 @@
...
+
+This proposal is to implement the WhatWG URL Standard by introducing a new
+`URL` *global*.
+
+The *existing* `url` module would **remain** is it currently is, with the
+addition of a single `require('url').URL` export that can be used an an
+alternative way of the getting a reference to the new `URL` constructor.
+
+Existing locations within core that currently accept a URL string or the
```
yorkie: Maybe we need to use the lower case "url" to stand for the url string to keep consistence with IDL source in https://url.spec.whatwg.org/

yorkie:
@jasnell How about the following 2 static methods which are defined at standard IDL:
- `domainToASCII(domain)`
- `domainToUnicode(domain)`

I didn't see those in the proposal :-(

jasnell:
Still considering those. They are easy enough to implement given the
punycode module but I'm not sure how extensively they're used.

jasnell:
There will also be other differences. For instance, I'm not sure if we need
the host parsing component.

targos:
>  I'm not sure yet if theirs is a pure JS impl or not. 

There is https://github.com/jsdom/whatwg-url

cc @domenic

jasnell:
Yeah, I'm familiar with (and use) `whatwg-url`. @domenic, how would you feel about the possibility of pulling the whatwg-url implementation into core? I'm not yet that familiar with how it is implemented internally but having a url standard compliant url parser built into core would be very good.

domenic:
Yep, that's @Sebmaster's most excellent work. It's not super optimized, but would be a good starting point.

This thread is very exciting and I'm glad there is an appetite for the idea!! The idea of a global, the same as in browsers, is great.

I think there are two separable problems here:
- Parsing URLs according to the URL Standard. This could possibly even be done for `require("url")` in a semver-major bump, as mostly it will change edge cases.
- Introducing a URL Standard-compliant API for URL manipulation. There are a couple things to note here:
  - The URL Standard uses the USVString type, which (per a convoluted series of specs) essentially means that unpaired surrogates get censored. This is generally a good property, and recommended by Unicode, as network boundaries do not expect unpaired surrogates. But it's not immediately obvious from the algorithms.
  - The URL Standard takes great pains to ensure that the URL is always consistent and usable, by using setters and getters. So e.g. if you do `url.pathname = "\uDC01"`, then that will initiate a parse in the pathname override state, and then if you trigger the getter, you'll find `url.pathname === "/%EF%BF%BD"`. Note how this both appends the leading slash, and also performs the USVString surrogate pair censoring (`decodeURIComponent("%EF%BF%BD").charCodeAt(0) === 0xFFFD`, not `0xDC01`). This is a fairly important property IMO.

There are tests of the URL Standard at https://github.com/w3c/web-platform-tests/tree/master/url, and whatwg-url has a runner. The coverage is pretty reasonable; see https://github.com/w3c/web-platform-tests/issues/3018

domenic:
> @domenic, how would you feel about the possibility of pulling the whatwg-url implementation into core?

Oh, that'd be very cool! I guess my only concern is that we weren't concerned with speed when writing it, so there is probably lots of low-hanging fruit for performance improvements.

You'd also need to do a bit of work to decouple it from webidl-conversions and webidl2js. If you `npm install` it you'll find that it follows the same impl/wrapper strategy as browsers do, where there's a "wrapper" that takes care of USVString conversion and so on, delegating to the "impl" where parsing occurs, which in turn delegates to the URL state machine code. At least one of these layers could be disintermediated, although perhaps benchmarking should be done to see if that's actually the area where most improvement could be made.

jasnell:
Ok, I'll dig in and explore the whatwg-url internals and see what can be done reasonably. Before getting too deep into this I'd definitely like to get more +1's from collaborators tho. I'd really like to see this happen tho.

jasnell:
> I think there are two separable problems here

Definitely agree that it's worth separating these. Not sure about modifying the existing `require('url')` too much tho -- ideally once the global `URL` is there for a while we could simply deprecate `require('url')` (and possibly even `require('querystring')`) entirely with an eye towards reducing the Node.js-specific API surface area.

Qard:
Well, you can have a 👍 from me! More browser/server unity would be great. 😺

Qard:
BTW, I like the idea of the global, since that's what browsers do, but I'd recommend _not_ exposing it in `require('url')` because then people might get used to it being there rather than available globally, making future removal of the module more difficult.

jasnell:
The only concern I would have with that, @Qard, is that if there is existing code that does `URL = whatever`, there would be no way of recovering the original global. That should be a limited edge case, however.

chrisdickinson:
This is not an argument for or against, but a request for more background. You've described what work you would like to do, and how you would do it — but I don't see much in the way of "why" we'd want to do this. What is not working about the status quo that we will rectify by adding a new global object? What value do Node users gain, concretely, from the URL object?

Qard:
Maybe store it on `process.URL` for safe keeping?

MylesBorins:
@chrisdickinson personally I see quite a bit of benefit in minimizing the delta between node + the browser as far as utility API's like this are concerned.

jasnell:
@chrisdickinson ... the why is straightforward: Currently Node.js' URL parsing has a number of issues in terms of not following the standardized behavior implemented by browsers. Examples of those failures can be seen in the test case I referenced [here](https://github.com/nodejs/node-eps/pull/28#issuecomment-223045821). There are also differences in the Node.js provided API that are largely unnecessary. This work would give us an opportunity to not only provide more robust URL parsing, but to provide a unified, non-Node.js specific API.

@Qard ... that's certainly a possibility

jasnell:
FWIW, looks like I mis-remembered the number of failures ;-) ... here's the exact count:

```
bash-3.2$ ./node test/known_issues/test-url-parse-conformance.js 
Unknown globals: URL

assert.js:90
  throw new assert.AssertionError({
  ^
AssertionError: 160 failed tests (out of 352)
    at Object.<anonymous> (/Users/james/Node/main/node/test/known_issues/test-url-parse-conformance.js:57:8)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Function.Module.runMain (module.js:575:10)
    at startup (node.js:152:18)
    at node.js:449:3
bash-3.2$ 
```

XXX-url.md
```diff
@@ -0,0 +1,75 @@
+| Title  | Implement WHATWG URL Spec   |
+|--------|-----------------------------|
+| Author | @jasnell                    |
+| Status | DRAFT                       |
+| Date   | 2016-06-01T09:00:00-07:00   |
+
+## Description
+
+The WhatWG URL Standard specifies updated syntax, parsing and serialization of
```
domenic: s/WhatWG/WHATWG/g

XXX-url.md
```diff
@@ -0,0 +1,75 @@
...
+This proposal is to implement the WhatWG URL Standard by introducing a new
+`URL` *global*.
+
+The *existing* `url` module would **remain** is it currently is, with the
+addition of a single `require('url').URL` export that can be used an an
+alternative way of the getting a reference to the new `URL` constructor.
+
+Existing locations within core that currently accept a URL string or the
+parsed results of `require('url').parse()` would be modified to also accept
+the `URL` object. Instances within core that current parse URL strings using
```
domenic: If possible I'd suggest not doing type-checking but instead just doing string coercion uniformly, assuming the existing url module has a reasonable toString() method.

XXX-url.md
```diff
@@ -0,0 +1,75 @@
...
+`require('url').parse()` would be modified to use the new `URL` global instead.
```
domenic: How would this impact http.request?

XXX-url.md
```diff
@@ -0,0 +1,75 @@
...
+* `href` is a `string` containing the URL to parse.
+* `base` is either a `string` or a `URL` that contains the base URL to resolve
+  against while parsing.
+
+See https://url.spec.whatwg.org/#urlutils-members for detail on the properties
+of the `URL` object.
+
+### `URLSearchParams`
+
+The `URLSearchParams` object is not intended to be creatable by users. It is
```
domenic: Per spec it is meant to be creatable by users...

XXX-url.md
jasnell: The existing URL module uses a `require('url').format()` method that serializes the "url" object that is returned by `require('url').parse()`. The returned URL object does not currently implement a reasonable toString()

```
> var u = url.parse('http://example.org');
undefined
> u.toString()
'[object Object]'
> url.format(u)
'http://example.org/'
> 
```

This is actually one of the existing API deficiencies that I'd like to see improved... A new API modeled after the browsers `URL` global would be a significant usability improvement.

XXX-url.md
jasnell: acknowledged. Will fix that in the next update.

XXX-url.md
domenic: Ah OK. Then one approach would be type-testing for the existing url module and string coercion for whatever's left, but maybe it's simpler just to do type-testing all the way.

joepie91:
My general dislike of constructors aside, I have to second @chrisdickinson's concern here. It's not clear to me why we'd want this as a global. Sure, that's how browsers implement it, but Node is not a browser, and browser APIs have historically been designed with a fundamentally different model in mind (namely, the lack of CommonJS).

This proposal, in its current form, is starting to look like it'll drag Node in the direction of PHP - endlessly adding APIs and not really trying to enforce or encourage consistency. As it is, this will just further confuse users as to whether they should be `require()`ing a module or expecting a global to be there - not to mention it being unclear whether they should use the `url` module or the `URL` global. Bundling tools _already_ take care of adding in 'fake' modules that point at browser APIs, so what exactly are we gaining here by trying to make it "more like browser APIs"?

As a separate question, how - if at all - does this cover the "insert object of URL components, receive stringified URL" usecase?

EDIT: Further question: Are there any major concerns about just fixing the existing `url` module in a next major (breaking) release? This would seem preferable to me, but I don't know if there are any major roadblocks that I might not be aware of, or whether this has been discussed somewhere before.

XXX-url.md
jasnell: ideally, there would be little outward impact other than being able to pass an instance of the global `URL` in as the argument... e.g.

``` js
const url = new URL('/foo', 'http://example.org/bar');
http.get(url, (res) => { /* ... */ });
```

XXX-url.md
jasnell: Ok. Can update that.

jasnell:
Let's separate the concerns just a bit. We _could_ implement this but _not_ make it a global, and just have it accessible via the existing `url` module (e.g. `const URL = require('url').URL`). We could go this route and _still_ eventually deprecate the Node.js specific API. I would be ok with going that route. This proposes it as a global to be consistent with browsers, but that's not critical if we cannot get consensus on it.

Second, on the point:

> endlessly adding APIs and not really trying to enforce or encourage consistency

My goal here is to eventually be able to fully deprecate the existing `url` and `querystring` modules with the hope of _reducing_ the Node.js specific API surface area. Obviously that's not something that would happen quickly, tho, so the concern is definitely noted.

> does this cover the "insert object of URL components, receive stringified URL" usecase?

It would in-so-far as the `URL` object as defined by the WHATWG spec includes `toString()` and the `href` property to provide serialization of the object. If what you're referring to is the ability to create a non-`URL` object and serializing it, that's not something that's currently supported by the WHATWG spec but it's something that can be easily maintained.

joepie91:
> We could implement this but not make it a global, and just have it accessible via the existing url module (e.g. const URL = require('url').URL). We could go this route and still eventually deprecate the Node.js specific API. I would be ok with going that route. This proposes it as a global to be consistent with browsers, but that's not critical if we cannot get consensus on it.

That seems like a more workable solution to me. There would still likely be short-term user confusion while both APIs exist, however, so this would be something that'd require very clear documentation. I realize that you might've missed the question I added to my previous post later on - could you have a look at that suggestion as well?

> My goal here is to eventually be able to fully deprecate the existing url and querystring modules with the hope of reducing the Node.js specific API surface area. Obviously that's not something that would happen quickly, tho, so the concern is definitely noted.

I understand that, but I still don't really see the value in trying to bring the Node.js API in line with browser APIs, as opposed to just keeping it internally consistent (given that Node.js was designed with CommonJS in mind, and switching to globals for everything would be a step back). In the end, Node.js _is_ its own environment, and I can't see how trying to reduce the API to the lowest common denominator would benefit usability.

> If what you're referring to is the ability to create a non-URL object a serializing it, that's not something that's currently supported by the WHATWG spec but it's something that can be easily maintained.

That's probably what I mean, yeah - I'm thinking of something along the lines of the `url.format` method, that can just directly accept a POJO.

alliejp-tv:
I believe it is valuable to implement the WhatWG URL Standard, but I don't understand why you would implement it as a global instead of simply fixing the `url` module. 

IMO @joepie91's concern is valid; I think we should stick to the existing accepted behavior. There's a reason Node.js is so easy to write, and that's because (in the vast majority of cases I have experienced), it _makes sense_, and it makes sense because it's consistent.

ChALkeR:
@jasnell 

This is the complete list of current properties of the global object that were not defined in the ECMA specs:
- `global`,
- `process`,
- `GLOBAL`, `root` — hard-deprecated aliases for `global`,
- `Buffer`,
- `clearImmediate`, `clearInterval`, `clearTimeout`, `setImmediate`, `setInterval`, `setTimeout`,
- `console`.

Almost all of those are there for a very good reason, and adding more should be considered very thoughtfully. I don't think that `URL` justifies that. We could add `fetch` next. And `XMLHttpRequest`. And so on — the list would be of indefinite length, once we start.

_Upd: for comparison, Chrome has more than 700 of those, not counting the ones defined by ECMA specs._

That said, I'm +1 on the idea of bringing `URL` to Node.js and perhaps eventually replacing Node.js specific api with it. And I think it's a great idea.

Not sure about exposing `URL` on `require('url')`, though, if the long-term plan is to deprecate `url`. Perhaps a different module name?

jasnell:
Ok, I think the building trend is to avoid the introduction of a new global, and that is fine. I'll pull that from the proposal. It can be revisited later on if necessary.

> instead of simply fixing the url module

My primary concern here is backwards compatibility. I don't simply want to modify the existing `url` module because doing so would likely breaking existing code. Also, fixing the existing code to be compliant with the WHATWG spec would be roughly the same amount of work as simply doing a clean new implementation. By implementing the `URL` object as a parallel API, we can provide a clear transition path off the Node.js specific API and onto the standard API, then fully deprecate and eventually remove the existing API. 

Done properly, the impact to existing code should be as minimal as possible, and if we are not introducing this as a global, then we would need to, at the very least, keep the `require('url')` module but eventually deprecate the Node.js specific APIs exposed by it. e.g.:

``` js
// deprecated APIs
const url = require('url');
url.parse();
url.format();
url.resolve();

// new API
const URL = require('url').URL;
new URL(url, base)
URL.format(urlLikeObject)
```

alliejp-tv:
@jasnell Alright, I'm more than comfortable with that.

jasnell:
In fact, thinking about it further, we may be able to simplify the transition even more by simply having the `url` module directly export the new `URL` object and hanging the deprecated `parse()` and `resolve()` methods as statics off of it. Existing code should be unaffected.

```
const URL = require('url');
new URL(url, base);
URL.format(urlLikeObject);

// deprecated API
URL.parse(); // the existing parse method
URL.resolve(); // the existing resolve method
```

(Anyone currently doing `const url = require('url')` would see no real difference)

domenic:
I'd caution against extending the standard URL with nonstandard methods, if the goal is to be able to help write cross-environment code. That includes `.format`.

jasnell:
@domenic ... noted ... then perhaps simply

``` js
const url = require('url');

// deprecated APIs
url.parse();
url.resolve();

// retained existing API
url.format(urlLikeObject);

// new API
const URL = require('url').URL;
new URL(url, base);
```

jasnell:
Updated based on the feedback

chrisdickinson:
@jasnell:

> ... the why is straightforward: Currently Node.js' URL parsing has a number of issues in terms of not following the standardized behavior implemented by browsers. Examples of those failures can be seen in the test case I referenced here. There are also differences in the Node.js provided API that are largely unnecessary. This work would give us an opportunity to not only provide more robust URL parsing, but to provide a unified, non-Node.js specific API.

Have the differences between WHATWG's URL spec and Node's caused errors _for users_? Issue links would help give the proposal concrete grounding.

> My goal here is to eventually be able to fully deprecate the existing url and querystring modules with the hope of reducing the Node.js specific API surface area. Obviously that's not something that would happen quickly, tho, so the concern is definitely noted.

This is notably something that we've tried before and have not been successful at; both in the larger sense of deprecation and API surface reduction, and in the specific sense that we've tried to replace the URL subsystem with a more spec-compliant version. What is different about this approach that we should expect a different outcome than before?

jasnell:
@chrisdickinson ... quick survey of URL related issues/prs ... 
- https://github.com/nodejs/node/issues/5832
- https://github.com/nodejs/node/issues/4101
- https://github.com/nodejs/node/issues/3361
- https://github.com/nodejs/node/pull/2937
- https://github.com/nodejs/node/issues/2277
- https://github.com/nodejs/node/issues/2114
- https://github.com/nodejs/node-v0.x-archive/pull/14146
- https://github.com/nodejs/node-v0.x-archive/issues/9404
- https://github.com/nodejs/node-v0.x-archive/issues/9117
- https://github.com/nodejs/node-v0.x-archive/issues/8725
- https://github.com/nodejs/node-v0.x-archive/issues/8636
- https://github.com/nodejs/node-v0.x-archive/issues/8520
- https://github.com/nodejs/node-v0.x-archive/issues/5452

The difference with this approach is that the existing `url` parse would not be replaced outright. To support existing users, it would be maintained and soft-deprecated at first, then hard deprecated in the next major beyond that, giving a clear transition path. Also, given that this moves towards using an API that many js developers are already familiar with, there is increased incentive to change.

chrisdickinson:
Thanks for the links.

Assuming it makes sense for us to address the problems folks are having with the `url` module, who maintains the new URL parser? Is this something that we will vendor from chromium? If so, how quickly does it change, & how easy is it to pull new versions? In the worst case, are we stuck maintaining two URL parsers? In the best case, do we only have the deprecated url parser, or no url parser?

> The difference with this approach is that the existing url parse would not be replaced outright. To support existing users, it would be maintained and soft-deprecated at first, then hard deprecated in the next major beyond that, giving a clear transition path.

If Node v9 came out with a hard deprecation (printed warning) on all `url` module usage, do you imagine we'd see less `url` use, or less adoption of Node v9?

chrisdickinson:
I agree with @TheAlphaNerd that there's definitely value in reducing the differences between browser APIs and Node APIs, but I want us to be clear about _what_ we're trying to solve, _why_ we're trying to solve it, and _how much_ we think it'll cost us.

Spec adherence for spec adherence's sake has some value, but not necessarily enough value to justify moving the ecosystem's cheese. Spec adherence as a side-effect of solving concrete problems our users are facing it has much more value. Does it have enough value for us to justify potentially maintaining two URL parsers into the future?

For my part, I lean towards "yes", but I'd prefer that information to be in the proposal so others can make that evaluation & refine our collective predictions about maintenance cost.

jasnell:
@chrisdickinson ... the implementation will most likely be a mix of green-field and stuff adopted/borrowed from the whatwg-url module. Now that I've had a chance to dig into the details of that module's code, there is quite a bit that would need to be optimized to get the necessary performance. What I would likely do is create a fork of that module and begin working on various optimizations. From there it can be adapted into core. Done, properly, the changes made to optimize performance could go back into the whatwg-url module also, allowing us to easily pull any modifications that are made there back in.

trevnorris:
-1 for global, and I don't like the idea of a constructor. For quite a while I've wanted to simply change node's existing parser to match the WhatWG spec, and release it on a semver-major. Not because I give a crap about complying with the browser, but because this feature has enough edge cases where it could be confused to be the same based on only a sample of URLs. Path of least surprise and all. If that's a no go then I can live with `require('url').URL`.

jasnell:
The path of least surprise would be to use the constructor since that's
exactly how things work on the browser side and is exactly what the WHATWG
spec defines.

Given enough time we could replace the `require('url').parse()` method to
follow the more compliant parsing but part of the while point of doing
things the way I've proposed is to avoid the semver-major breaking change
in the near term in favor of a more incremental approach (that is, add
`URL` in v6; soft deprecate `parse()` in v7; with hard deprecation in v8).

trevnorris:
Yeah. I recede my point about not wanting a constructor. Since `parse()` returns an object fundamentally things wouldn't change. I'm cool with that. Still no global though. And I do think that instead of deprecating `.parse()` it would be better to just move to the same parser as `URL()`.

ChALkeR:
@jasnell I am not sure if hard deprecation in v8 would be achievable.

jasnell:
Possibly (likely) not. As I said, it's a goal :). Realistically it would
likely take longer unless we decided to simply flip the switch as
@trevnorris suggests

Jxck:
questions.
1. whatwg url spec has URLSearchParams which duplicate role with querystring module. querystring will leave as is ?
2. whatwg url spec depends on encoding spec too. (https://encoding.spec.whatwg.org/) will Encoding be separate modules like `require('encodings')` ? or internal only ?

jasnell:
The querystring module would remain unchanged. URLSearchParams would be exposed only via the URL class. It's implementation could easily be backed by the querystring module, in fact (I've already done this actually).

No part of the encodings spec would be exported or visibly implemented. Essentially, in terms of new API surface, this would only export the URL and URLSearchParams objects.

Qard:
I'm okay with non-global. It's likely lots of people would use a polyfill module anyway, which does the native-or-polyfill return type thing, for wider support.

yosuke-furukawa:
I am a bit concerned about schema. 
According to [spec](https://url.spec.whatwg.org/#special-scheme), whatwg url handles the following schema,
- `ftp`
- `http`
- `https`
- `file`
- `gopher`
- `ws`
- `wss`

But Node.js needs to handle other schema like `git`...

I tried but I got the following results.

in chrome

```
> new URL('git://github.com/foo/bar');
host:""
hostname:""
href:"git://github.com/bar/buz"
origin:"git://"
pathname:"//github.com/bar/buz"
port:""
protocol:"git:"
```

in node

```
> url.parse('git://github.com/foo/bar');
protocol: 'git:',
host: 'github.com',
port: null,
hostname: 'github.com',
hash: null,
search: null,
query: null,
pathname: '/bar/buz',
path: '/bar/buz',
href: 'git://github.com/bar/buz'
```

Jxck:
FYI. I worked on same topic years ago (in typescript because of following WebIDL static interface)
the goal was implement perfect whatwg fetch, includes perfect whatwg url.

if you will implement url, you requires thease.
- obtain-unicode: https://github.com/Jxck/obtain-unicode
- encoding: https://github.com/Jxck/utf8-encoding (I've done utf8 only)
- url-searchparams: https://github.com/Jxck/URLSearchParams
- url: https://github.com/Jxck/URL/blob/master/url.ts

most difficult thing is _domain to asci_ and _domain to unicode_. http://www.unicode.org/reports/tr46/#ToASCII
I gave up this point :(

jasnell:
Yeah, it will be necessary to expand the list of "special" schemes
supported by the parser. This shouldn't be that difficult to do.

yosuke-furukawa:
my concern is clear, sounds good.

rvagg:
@jasnell I know you're resisting this because of the work, but I'm with the folks in here that would like a more concrete sense of what the difference is between our impl and the spec. I know you have a test case for it and can count the number of tests but having a clearer picture of what changes would be required to shift the current impl to be spec compliant would likely help the decision. Perhaps it's waaaay to much breakage and in that case it rules out a semver-major break, or perhaps it's edge-casey breakage as Domenic suggests and we'd all be comfortable with shifting the impl in a semver-major and communicating it clearly while we do it.

Re increasing API surface area, I'm very skeptical that if we were to implement a second URL parser that we could easily get rid of the original and that we'll be stuck in an awkward limbo of having to support both. I'm not at all comfortable giving a +1 to doing this unless we have clear data ruling out other options.

jasnell:
@rvagg ... to be honest, I'd prefer to just replace the existing URL parser outright as a semver-major change, but I'm trying to be sensitive to the possibility of breaking things. Recent experiences around realpath, the module loader and stdio, as well as the breakage that occurred the last time significant changes to the existing url parser were attempted, would imply that being _overly_ sensitive to _any_ breaking change is warranted. In other words, even if breaking changes to the existing parser are minimal, I'd like to avoid making such breaking changes at all and instead provide an alternative implementation and API that users can transition to. So if the only option in core is to modify the existing parser in a non-backwards compatible way, I'd rather simply opt to iterate on whatwg-url as a userland module and recommend that developers use that as an alternative to the built in parser.

domenic:
> whatwg url spec depends on encoding spec too. (https://encoding.spec.whatwg.org/) will Encoding be separate modules like require('encodings') ? or internal only ?

There will be no real dependency for a Node implementation. It only depends on non-UTF-8 encodings when the "encoding override" parameter is used, which is only the case for web content.

tr46 is a real dependency for host parsing (and domainToUnicode/domainToASCII) and might be a real concern. https://github.com/Sebmaster/tr46.js/blob/master/index.js but also note that it includes a mapping table generated from IDNA.

> Yeah, it will be necessary to expand the list of "special" schemes supported by the parser. This shouldn't be that difficult to do.

I would be against this. Browsers don't parse special schemes this way, and we should not either. `git` should be parsed the same as `data`, I believe.

>  I know you're resisting this because of the work, but I'm with the folks in here that would like a more concrete sense of what the difference is between our impl and the spec. I know you have a test case for it and can count the number of tests but having a clearer picture of what changes would be required to shift the current impl to be spec compliant would likely help the decision.

I believe the differences in parsing are edge-casey, but the differences in API surface are extreme, because of the use of getters and setters. That is part of what sunk the old attempt to rewrite the URL parser, and that was only to use getters for for speed, without even implementing setters for consistency. I think the API surface of the URL Standard is really nice, personally, and think it would be great for core to adopt it.

To head off any suggestions of having core implement the parser but leaving the API surface to userland, this isn't really feasible unfortunately, as the API surface requires a lot of direct hooks into the middle of the parser to implement e.g. the setters that keep things consistent. The full low-level API surface necessary to implement a higher API is given in https://github.com/jsdom/whatwg-url#low-level-url-standard-api; as you can see it's a lot more than a simple string -> URL record parser (with the state overrides alone adding a lot of entry points).

trevnorris:
> I would be against this. Browsers don't parse special schemes this way, and we should not either. git should be parsed the same as data, I believe.

Overall I agree with you, but I believe this point also violates the path of least surprise. If users can't parse the additional protocols then they're unlikely to ever adopt the new API.

jasnell:
Just ran some tests. The whatwg-url parser handles git URLs with no
problems. The chrome parser is the one with the issue on it.

ben-page:
This is great. After writing the little PR above, I wrote my own implementation of the whatwg-url parser. The current parser is in adequate many ways and I think people expect more parity with browser's implementations.

I haven't released my implementation, because it doesn't do URL serializing or equivalence. Also, I wasn't sure how to drum up interest in such a significant change to Node.js. Anyway, I'm be interested in help with this in anyway I can.

ben-page:
On subject of scheme, a valid scheme can contain ASCII alphanumeric characters, `+`, `-`, or `.`. So parsing git schemes should not be a problem. The standard defines special schemes: `ftp`, `file`, `gopher`, `http`, `https`, `ws`, `wss`. But this list only exists to define the default IP port for these schemes. The one exception is the `file` scheme. It's parsed completely differently than other schemes.

jasnell:
EPS proposal updated. /cc @nodejs/ctc

XXX-url.md
```diff
@@ -0,0 +1,170 @@
...
+the URL standard and fails to pass 160 of the WHATWG URL parsing tests.
+
+This proposal is to implement the WHATWG URL Standard by modifying the existing
+`url` module to provide an implementation of the `URL` object and associated
+APIs. Doing so improves the robustness of URL parsing, provides consistency
+with browser js code, and can eventually allow the reduction of node.js specific
+APIs.
+
+Initially, the implementation would be introduced as an undocumented
+experimental feature exposed via a new `URL` property in the `url` module.
```
chrisdickinson: There's already a `Url` property that points to the old constructor/parser — this is likely to cause confusion. _If we go down the route of supporting a new URL parser_, we may as well expose it as a global. Code written for the browser expects it to be a global, we may as well avoid artificially increasing the delta between browser JS and server JS.

XXX-url.md
```diff
@@ -0,0 +1,170 @@
...
+
+Because the existing `require('url')` module remains, there should be no
+backwards compatibility concerns. Once a decision is made to provide official,
+documented support for the new URL implementation, a decision would be made
+about whether to switch the internal uses of `require('url').parse()` to the
+new `URL` implementation.
```
chrisdickinson: By leaving this decision to the future the proposal implicitly indicates that we'll be supporting two URL parsers indefinitely. The question for the CTC: is the value of a standard URL parser to our users worth the cost of us maintaining two URL parsers?

jasnell: The idea is to provide a smooth transition while the performance of the
standard parser is improved. My preference would be to deprecate and
eventually remove the existing url.parse.

XXX-url.md
jasnell: I'm all for it being a global.

trevnorris: I'm also fine with that. Already have `encodeURI()`, etc.

XXX-url.md
trevnorris: Would it be enough to say something like "once performance is on par, the old parser will go through a normal depreciation cycle"?

jasnell: Works for me

chrisdickinson: My preference would for us to say "url.parse will not be deprecated, but the docs for the module will include an admonition to use the global `URL` object, linking to the MDN docs."

Trott: @chrisdickinson I don't want to get into hair-splitting over terminology, but I'm curious what the difference is between what you are describing and "deprecated" since "deprecated" literally means to recommend against something. Is it that too many people will (incorrectly) take "deprecated" to mean "to be removed at some point in the future"?

jokeyrhyme: I've had a few cases in recent years where I have read "deprecated", and understood its literal dictionary definition, only to find that the author(s) actually meant "removed".

Perhaps we should always make that word a hyperlink to its dictionary / engineering definition?

trevnorris: How about we revisit our wording, because "deprecate" continues to bring grief. So how about APIs that are deprecated mean that we suggest not using it. Possibly also that it gets lower priority with patches.  Then something like "removable" to mean that it's on the roadmap to be removed and there's little sense in attempting to improve it.

chrisdickinson: @Trott, @jokeyrhyme: There are multiple levels of deprecation — there's "admonition against use in the docs" deprecation, and there's "actively emits warnings when used" deprecation. Deprecating (-as-in-warning) and removing the old url parser is probably unrealistic given the number of users it would affect.

jasnell: https://github.com/nodejs/node/issues/4677

jasnell:
Putting this on the ctc-agenda as I believe it's ready for discussion/review.

evanlucas:
> I would be against this. Browsers don't parse special schemes this way, and we should not either. git should be parsed the same as data, I believe.

Does npm rely on the parsing of urls with support for the `git` protocol? If so, do we have a migration path for that?

trevnorris:
There's a certain loss of functionality by using `new URL()` over `url.parse()`. Specifically that `url.parse()` isn't just for URLs but also for URIs. For example: `url.parse('github.com')` works fine but `new URL('github.com')` throws.

Taking the following excerpt from the [WHATWG URL specification](https://url.spec.whatwg.org/):

> Standardize on the term URL. URI and IRI are just confusing. In practice a single algorithm is used for both so keeping them distinct is not helping anyone. URL also easily wins the search result popularity contest.

It seems that `URL` is not meant to mean _only_ URL. Even though there a scheme is always necessary (whether absolute or relative URL). Because of this any scheme is acceptable, but possibly parsed as expected. For example:

```
new URL('git://github.com/foo/bar').pathname == '//github.com/foo/bar';
```

Even though `git:` has been [provisionally registered](http://www.iana.org/assignments/uri-schemes/uri-schemes.txt) with the IANA it is still not parsed as a standard domain. It should be possible to properly properly parse the `pathname` from the above.

Taking from [RFC3986](https://www.ietf.org/rfc/rfc3986.txt) we can see a simple generic way of parsing the URL that I think could be used at minimum to parse unfamiliar schemes:

```
         foo://example.com:8042/over/there?name=ferret#nose
         \_/   \______________/\_________/ \_________/ \__/
          |           |            |            |        |
       scheme     authority       path        query   fragment
          |   _____________________|__
         / \ /                        \
         urn:example:animal:ferret:nose
```

@domenic 

> I would be against this. Browsers don't parse special schemes this way, and we should not either. `git` should be parsed the same as `data`, I believe.

Not to be rude, but check your own work first. Chrome has added an exception to `URL()` for its own needs:

```
new URL('chrome-extension://adlfkjaslkfj/foo/bar.html').pathname === '/foo/bar.html';
```

Additionally, `URL()` today doesn't parse unknown schemes as `data`:

```
const u = new URL('git://github.com/foo/bar');
u.protocol === 'git:';
u.origin === 'git://';
```

All this said, we should support other/unknown schemes properly with `URL()`.

domenic:
It's already been clarified upthread that I was incorrect about special schemes. There are no special schemes besides file:

trevnorris:
@domenic oop. sorry about that. completely glanced over that comment.

jasnell:
Adding git as a special scheme would be trivial if necessary. If necessary
we should first try to upstream such a change into the URL spec itself.

jasnell:
After further testing, git URLs are handled by this implementation without any problems whatsoever using the implementation in https://github.com/nodejs/node-eps/pull/28.

jasnell:
The CTC discussed this today and has decided to advance this to DRAFT status with the exception of the `URL` constructor not being a global. I will get the PR updated to remove part about it being a global and will get it landed. Thank you all.

yorkie:
By the way, after the core implemented the WHATWG URL standard, I think also should start deprecating the `path` in the `http` module at [here](https://nodejs.org/dist/latest-v6.x/docs/api/http.html#http_http_request_options_callback) where we were using the `path` in request option.

jasnell:
@yorkie ... eventually, once the new URL object is no longer experimental, my goal is to replace use of the existing url.parse within the http module entirely. That's not going to be easy tho. It's going to take a long time.

jasnell:
Landed in f16f770

PR #32, opened 2016-06-03
uv_link_t: initial

indutny:
cc @nodejs/ctc @trevnorris @nodejs/collaborators @saghul

### DIFF
--- a/xxx-uv_link_t.md
+++ b/xxx-uv_link_t.md
@@ -38,9 +38,8 @@ From [`uv_link_t`][0] readme:
     high-performant way using very narrow interfaces.
 
 Given that [`uv_link_t`][0] depends only on [libuv][1], it is very easy to
-envision how it will be used in Node.js addons. Even core modules will no longer
-be required to be distributed with core, since they could be built without core
-dependencies.
+envision how it will be used in Node.js addons. Even core modules will could
+be reimplemented on a user level without a need to patch the core itself.
 
 Additionally, with the `uv_link_observer_t` (which is a part of [`uv_link_t`][0]
 distribution), all of the reads that right now happen internally within the
--- a/xxx-uv_link_t.md
+++ b/xxx-uv_link_t.md
@@ -13,7 +13,7 @@ classes that inherit from `StreamBase`.
 
 The fast HTTP and TLS protocol implementation in Node core depends on so called
 `StreamBase` C++ class. This class is an analog of JavaScript streams in C++.
-The main ideas behind `StreamBase` is to:
+The main ideas behind `StreamBase` are to:
 
 1. Avoid unnecessary memory allocations by reading data directly into buffer
    from which it will be synchronously parsed
@@ -28,12 +28,12 @@ From [`uv_link_t`][0] readme:
 
     Chainable libuv streams.
 
-    It is quite easy to write a TCP server/client in [libuv][0]. Writing HTTP
+    It is quite easy to write a TCP server/client in [libuv][1]. Writing HTTP
     server/client is a bit harder. Writing HTTP server/client on top of TLS
     server could be unwieldy.
 
     `uv_link_t` aims to solve complexity problem that quickly escalates once
-    using multiple layers of protocols in [libuv][0] by providing a way to
+    using multiple layers of protocols in [libuv][1] by providing a way to
     implement protocols separately and chain them together in easy and
     high-performant way using very narrow interfaces.
 
@@ -57,8 +57,8 @@ They can be combined together quite trivially as demonstrated in
 
 Another abstraction level could be added for multiplexing [`uv_link_t`][0]s to
 provide a common C interface for http2 and http1.1. This will help us bring
-a http2 implementation into the core while reusing as much of the existing code
-as possible.
+a http2 implementation into the core while reusing as much as possible of the
+existing code.
 
 [0]: https://github.com/indutny/uv_link_t
 [1]: https://github.com/libuv/libuv
--- a/xxx-uv_link_t.md
+++ b/xxx-uv_link_t.md
@@ -38,7 +38,7 @@ From [`uv_link_t`][0] readme:
     high-performant way using very narrow interfaces.
 
 Given that [`uv_link_t`][0] depends only on [libuv][1], it is very easy to
-envision how it will be used in Node.js addons. Even core modules will could
+envision how it will be used in Node.js addons. Even core modules could
 be reimplemented on a user level without a need to patch the core itself.
 
 Additionally, with the `uv_link_observer_t` (which is a part of [`uv_link_t`][0]
--- a/xxx-uv_link_t.md
+++ b/xxx-uv_link_t.md
@@ -12,7 +12,8 @@ classes that inherit from `StreamBase`.
 ## 2. Current state of C++ Streams in Node
 
 The fast HTTP and TLS protocol implementation in Node core depends on so called
-`StreamBase` C++ class. This class is an analog of JavaScript streams in C++.
+`StreamBase` C++ class and auxiliary `StreamReq` and `StreamResources` classes.
+This class is an analog of JavaScript streams in C++.
 The main ideas behind `StreamBase` are to:
 
 1. Avoid unnecessary memory allocations by reading data directly into buffer
@@ -34,7 +35,7 @@ From [`uv_link_t`][0] readme:
 
     `uv_link_t` aims to solve complexity problem that quickly escalates once
     using multiple layers of protocols in [libuv][1] by providing a way to
-    implement protocols separately and chain them together in easy and
+    implement protocols separately and chain them together in an easy and
     high-performant way using very narrow interfaces.
 
 Given that [`uv_link_t`][0] depends only on [libuv][1], it is very easy to
--- a/xxx-uv_link_t.md
+++ b/xxx-uv_link_t.md
@@ -42,6 +42,12 @@ envision how it will be used in Node.js addons. Even core modules will no longer
 be required to be distributed with core, since they could be built without core
 dependencies.
 
+Additionally, with the `uv_link_observer_t` (which is a part of [`uv_link_t`][0]
+distribution), all of the reads that right now happen internally within the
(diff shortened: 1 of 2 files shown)


eljefedelrodeodeljefe:
As far as I can see, this would tremendously help native addons writers also, since StreamBase is indeed very bulky. Decreasing boilerplate one has to implement to hook in C programs into js-streams is definitely worth an addition like this. Awesome work, btw.

Qard:
Indeed, this looks great! The more we can leverage libuv to do the heavy lifting generically, the better. :smile_cat:

xxx-uv_link_t.md
```diff
@@ -0,0 +1,67 @@
...
+## 1. Proposal
+
+I propose to replace `StreamBase` with [`uv_link_t`][0], and port existing
+classes that inherit from `StreamBase`.
+
+## 2. Current state of C++ Streams in Node
+
+The fast HTTP and TLS protocol implementation in Node core depends on so called
+`StreamBase` C++ class. This class is an analog of JavaScript streams in C++.
+The main ideas behind `StreamBase` is to:
```
thefourtheye: nit: are to

xxx-uv_link_t.md
```diff
@@ -0,0 +1,67 @@
...
+* [`uv_http_t`][3]
+
+They can be combined together quite trivially as demonstrated in
+[file-shooter][4].
+
+## 5. Further Thoughts
+
+Another abstraction level could be added for multiplexing [`uv_link_t`][0]s to
+provide a common C interface for http2 and http1.1. This will help us bring
+a http2 implementation into the core while reusing as much of the existing code
```
thefourtheye: as much as?

xxx-uv_link_t.md
```diff
@@ -0,0 +1,67 @@
...
+From [`uv_link_t`][0] readme:
+
+    Chainable libuv streams.
+
+    It is quite easy to write a TCP server/client in [libuv][0]. Writing HTTP
+    server/client is a bit harder. Writing HTTP server/client on top of TLS
+    server could be unwieldy.
+
+    `uv_link_t` aims to solve complexity problem that quickly escalates once
+    using multiple layers of protocols in [libuv][0] by providing a way to
```
thefourtheye: Did you mean `[libuv][1]`?

jasnell:
oooo... yes please. this definitely looks promising.

xxx-uv_link_t.md
indutny: Ack.

xxx-uv_link_t.md
indutny: Ack, rephrased.

xxx-uv_link_t.md
indutny: Ack.

xxx-uv_link_t.md
```diff
@@ -0,0 +1,67 @@
+| Title  | Port core to uv_link_t      |
+|--------|-----------------------------|
+| Author | @indutny                    |
+| Status | DRAFT                       |
+| Date   | 2016-06-03                  |
+
+## 1. Proposal
+
+I propose to replace `StreamBase` with [`uv_link_t`][0], and port existing
+classes that inherit from `StreamBase`.
```
trevnorris, line 10: We had a conflict in `StreamBase` that lead to the method `GetAsyncWrap()` and `GetObject()`. Will this play nicely with `AsyncWrap`, and allow the class to propagate like it does with `JSStream`?

xxx-uv_link_t.md
```diff
@@ -0,0 +1,67 @@
...
+
+## 2. Current state of C++ Streams in Node
+
+The fast HTTP and TLS protocol implementation in Node core depends on so called
+`StreamBase` C++ class. This class is an analog of JavaScript streams in C++.
+The main ideas behind `StreamBase` are to:
```
trevnorris, line 17: Does this mean you'll be removing associated classes like `StreamReq` and `StreamResource`?

xxx-uv_link_t.md
indutny, line 10: The conflict was due to the use of `vtable` and `virtual` methods in `StreamBase`. `uv_link_t` is a C project, there won't be any such conflict.

In fact, I believe that exposing an External getter for the pointer to internal `uv_link_t` is enough to get things running. This getter can be a per-class, or could be shared, but the inheritance won't be a roadblock for this.

xxx-uv_link_t.md
indutny, line 17: Yes.

indutny:
Added note about observability and `data` events. This is very important too.

xxx-uv_link_t.md
```diff
@@ -0,0 +1,73 @@
...
+    server could be unwieldy.
+
+    `uv_link_t` aims to solve complexity problem that quickly escalates once
+    using multiple layers of protocols in [libuv][1] by providing a way to
+    implement protocols separately and chain them together in easy and
+    high-performant way using very narrow interfaces.
+
+Given that [`uv_link_t`][0] depends only on [libuv][1], it is very easy to
+envision how it will be used in Node.js addons. Even core modules will no longer
+be required to be distributed with core, since they could be built without core
```
trevnorris: I have a hint of a concern that people will take this as meaning core modules will be distributed without core. Possibly change the wording to simply say that it'd be easier to write alternative implementations of core modules (e.g. alternative http implementations). Only suggest this to keep the proposal closer to core today (as an easier way for this EP to get accepted).

note: i'm not suggesting that i wouldn't like to distribute the core modules separately in the future. :)

trevnorris:
Overall I dig it. Would like @bnoordhuis' opinion.

xxx-uv_link_t.md
indutny: Ack, PTAL.

indutny:
@trevnorris thank you!

xxx-uv_link_t.md
Qard: Actually, I'm wondering about the idea of prototyping this in userland, since `uv_link_t`, `uv_http_t`, and `uv_ssl_t` are all fully separate from libuv itself currently. Can we make totally separate modules for those three parts and have them interact cleanly? That'd be a great proof of modularity. It could be very interesting to see lots of little uv extensions like this exposed as userland modules, a `uv_websocket_t` link comes to mind. 😸

indutny: This sounds lovely, @Qard . Though, we will still need to have access to the `uv_stream_t` which is currently hidden in core.

Qard: Maybe we could add an internal/undocumented function somewhere to return a `v8::External` containing the `uv_stream_t` to experiment with the idea?

indutny: @Qard do you think this should be a part of this proposal?

Qard: Not sure yet. It's more just an idea in my head at this point.

It might make acceptance of this EPS a little smoother though. We could just start with a single hidden function and iterate in userland for a bit. It'd provide the opportunity to develop a more practical understanding of the potential implementation than just bikeshedding a bunch in here.

trevnorris: I agree with @indutny. This is an idea that has a lot of potential and lots of ways we can be extended. Think it will be good to be strict about not allowing these things to creep so the initial implementation can land. At that point I think it'll be much easier to begin prototyping new ways of doing things.

xxx-uv_link_t.md
```diff
@@ -38,9 +38,8 @@ From [`uv_link_t`][0] readme:
     high-performant way using very narrow interfaces.
 
 Given that [`uv_link_t`][0] depends only on [libuv][1], it is very easy to
-envision how it will be used in Node.js addons. Even core modules will no longer
-be required to be distributed with core, since they could be built without core
-dependencies.
+envision how it will be used in Node.js addons. Even core modules will could
```
trevnorris: "will could" -> "could". Other than that minor thing, reads great. Thanks!

indutny: Ack.

indutny:
Thank you for review @trevnorris !

xxx-uv_link_t.md
rvagg, line 17: should be noted in the text

xxx-uv_link_t.md
```diff
@@ -0,0 +1,72 @@
...
+
+    Chainable libuv streams.
+
+    It is quite easy to write a TCP server/client in [libuv][1]. Writing HTTP
+    server/client is a bit harder. Writing HTTP server/client on top of TLS
+    server could be unwieldy.
+
+    `uv_link_t` aims to solve complexity problem that quickly escalates once
+    using multiple layers of protocols in [libuv][1] by providing a way to
+    implement protocols separately and chain them together in easy and
```
rvagg: "in _an_ easy and..."

bnoordhuis:
The document should provide some detail on how uv_link_t is supposed to work.  It currently just, ah, links to https://github.com/indutny/uv_link_t without further explanation.

xxx-uv_link_t.md
indutny, line 17: Ack.

xxx-uv_link_t.md
indutny: Ack.

indutny:
@bnoordhuis I can copy-paste "How?" from https://github.com/indutny/uv_link_t#how , will it suffice?

bnoordhuis:
Yes, but it should probably also say a few words about how it's going to be hooked up in node.js.

indutny:
@bnoordhuis sorry for delay, pushed update.

indutny:
Should I land this? Its been awhile since the last review, and I believe that there is a general agreement about this feature.

Trott:
> Should I land this?

Since it says status is DRAFT, and the README says all EPs regardless of status should be landed...I'd say yes. Landing it as DRAFT status especially should be uncontroversial.

indutny:
Alright, landing!

indutny:
Landed in 2c45bf0, thank you!

PR #37, opened 2016-07-29
Editorial fixes / suggestions for ES6 Module Interoperability

jmm:
Separate commits to make it easier to consider the changes individually.

56faad9f182df68e8c25ee86196100e7b414e155
Some of the changes are subjective style suggestions, but I think the _named export_ language here could be confusing in terms of ability to reassign the value of a binding inside the module that exports it:

> Attempting to reassign any named export will not work

Properties of namespace objects are live bindings, right?

f753479ec1856e347b7ff9dbcbe110ef37b9f139
Style nit. The formatting _ModuleNamespace_ or _ModuleNamespaceObject_ doesn't seem to appear in the spec. It seems to always be _Module Namespace [Exotic] Object_ (though sometimes it's lowercase and sometimes title case).

If you don't like the style changes, ¯\_(ツ)_/¯, but hopefully you find the other stuff worthwhile.

### DIFF
--- a/002-es6-modules.md
+++ b/002-es6-modules.md
@@ -548,7 +548,7 @@ No existing code will be affected.
 
 #### 5.6.1. ES exports are read only
 
-The objects create by an ES module are [ModuleNamespace Objects][5].
+The objects created by an ES module are [ModuleNamespace Objects][5].
 
 These have `[[Set]]` be a no-op and are read only views of the exports of an ES
 module. Attempting to reassign any named export will not work, but assigning to
--- a/002-es6-modules.md
+++ b/002-es6-modules.md
@@ -37,7 +37,7 @@ related syntax, and introduces:
     - Defines the list of imports via `[[ImportEntry]]`.
     - Defines the list of exports via `[[ExportEntry]]`.
 
-* **[ModuleNamespace]
+* **[Module Namespace]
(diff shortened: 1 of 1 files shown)


trevnorris:
cc @bmeck

bmeck:
after tc39 meeting there will be rather drastic changes, will review after a meeting on Wednesday about things

002-es6-modules.md
```diff
@@ -619,8 +625,8 @@ console.log(namespace.foo); // undefined
 
 // mutate to show 'yo' still exists as a binding
 import cjs_exports from './bad-cjs.js';
-cjs_exports.yo = 'lo again';
-console.log(namespace.yo); // 'yolo again'
+cjs_exports.yo += ' again';
+console.log(namespace.yo); // 'lo again'
```
sheerun, line 629: "yo again"?

jmm, line 629: Do you mean you think that would make a better example, or you think that's what the output should be? In both the original and updated code the output would be `lo again`, right? Personally I would probably just make the example `original` / `updated` or something like that.

Trott:
Thanks for the pull request! Modules work is now happening elsewhere (in the nodejs/modules and nodejs/node repositories) and this repository seems likely to be archived soon. I'm going to close this, but if you want these changes to land, by all means, comment and rebase.

PR #39, opened 2016-08-26
Rewrite 002 - esm

bmeck:
some discussions and problems with supporting various behaviors on a VM level, races with browsers, and upcoming spec changes have led to a drastic change in direction for the interop bridge.

https://gist.github.com/bmeck/52ee45e7c34d1eac44ce8c5fe436d753 has some relevant notes

notable:
- back to file extension
- more explicitly stated that ESM will not respect require.extensions in the algorithm
- `import` is url based
  - if passed full URL, no path searching is performed
  - else if starts with `../` , `./`, or `/` use as base URL against current script and search
  - else search in `node_modules`
    - added guards for potential web compat around `../` escaping `node_modules`
- ESM cache won't be shared / placed in require.cache at all
- ESM and `import` **ALWAYS** unwind stack prior to any evaluation
  - `import('old.js')` (spec being written) would always unwind the stack prior to evaluating.
- require of ESM returns `Promise<ModuleNamespace>` always, never synchronous
- no named imports from current modules (JSON / C++ / Node CommonJS)
  - all produce a single `default` export
- changes to what module scoped variables
  - removed `__filename`, `__dirname`, `require`, `module`, `exports`, `arguments`
  - link to tc39 spec on `this` value


002-es-modules.md
```diff
@@ -0,0 +1,525 @@
...
+foo = null;
+```
+
+```javascript
+// NCJS.js
+const esm = require('./es');
+// esm ~= Promise => {
+//   default => result_from_evaluating_foo;
+// }
+es_namespace.then(
```
Jessidhia: `esm.then`?

002-es-modules.md
```diff
@@ -0,0 +1,525 @@
...
+
+| Variable | Exists | Value |
+| ---- | ---- | --- |
+| this | y | [`undefined`](https://tc39.github.io/ecma262/#sec-module-environment-records-getthisbinding) |
+| arguments | n | |
+| require | n | |
+| module | n | |
+| exports | n | |
+| __filename | n | |
+| __dirname | n | |
```
Jessidhia, line 299: They will no longer exist as barewords, but they will still exist with a per-module value in `global`? (according to the "transpilers:" comment section in the gist)

If that is the case, wouldn't they (at least `__filename` and `__dirname`) only be current if read at the top level? This would require transpilers to inject a hoisted `const __filename = global.__filename, __dirname = global.__dirname;` if they are used.

Jessidhia: Ah, I misunderstood; they will actually not exist anymore and thus should become an attempt to read against `global` (which will probably result in `undefined`).

I suppose that, until `import.context.url` is available, a workaround to, at least, have _some_ known dirname, would be to have a NCJS module that just does `module.exports = __dirname`.

002-es-modules.md
bmeck: good catch

002-es-modules.md
bmeck: added note and workaround blurb

unional:
What about transitivity between the two?

bmeck:
@unional it will not be possible unfortunately.

unional:
cc/ @blakeembrey

002-es-modules.md
```diff
@@ -261,7 +261,21 @@ ESM will not be bootstrapped with magic variables and will await upcoming specif
...
+Although heavily advised against, you can have a NM module sibling for your ESM that can export these things:
+
+```js
+// expose.js
+module.exports = {__dirname};
+```
+
+```js
+// use.mjs
+import {__dirname} from './expose.js';
```
Jessidhia: Can't pluck `__dirname` from the `default` export :D

``` js
module.exports = __dirname;
```

``` js
import __dirname from './expose.js'
```

002-es-modules.md
```diff
@@ -0,0 +1,540 @@
...
+2. let `export` be the value of `module.exports`
+3. if there was an error, place the error in the ESM cache and return
+4. create an ESM with `{default:export}` as its namespace
+5. place the ESM in the ESM cache
+
+### 3.5.2. ESM to NCJS
+
+ESM are *never* placed in the NCJS cache (`require.cache`).
+
+If `require` resolves to an ESM source text, the following steps are taken.
```
hybrist: Could this be turned into `require.import`? It seems weird if the same function now sometimes returns a promise to a module and sometimes the exports directly. It seems unlikely that the same consumer would "gracefully" handle that switch between node versions, potentially leading to breakage on update.

bmeck: ESM always generate `Promise<ModuleNamespace>`s, NCJS always generate the value of `module.exports`

Don't really see the need to make a new function.

bmeck: A note, literally cannot form the ModuleNamespace synchronously all of the time; so we can't make `require(esm)` return a module namespace like babel currently does.

002-es-modules.md
```diff
@@ -0,0 +1,540 @@
+| Title  | ESM Interoperability |
+|--------|-----------------------------|
+| Author | @bmeck                      |
+| Status | DRAFT                       |
+| Date   | 2016-01-07                  |
```
MylesBorins: should the date be updated?

bmeck: maybe, unclear on this. Is it the date when things started or for last update / errata

002-es-modules.md
hybrist: Sure, but `loader.import` / `import` will give consistent return types for both.

``` js
// Both return promises!
loader.import('./es6.mjs').then(exports => {});
loader.import('./common.js').then(exports => {});

// Both "return" actual values!
import theActualExportES6 from './es6.mjs';
import theActualExportCommonJS from './common.js';

// require - suddenly the difference shows
require('./es6.mjs').then(exports => {}); // ah! works just like loader.import
require('./common.js').then(exports => {}); // oops, not a promise
```

This seems needlessly confusing to me. But maybe that's just personal preference. :)

bmeck: We could just ban `require` from loading ESM, I have thought about this idea a lot; though that cannot currently be shipped until standards form a real `import()` (can't be a property of a global due to needing context like what the current script location is)

hybrist: Proposal would be to add a `require.import` that always returns a promise and not change the semantics of `require` at all:

``` js
// Both return promises!
loader.import('./es6.mjs').then(exports => {});
loader.import('./common.js').then(exports => {});

// Both "return" actual values!
import theActualExportES6 from './es6.mjs';
import theActualExportCommonJS from './common.js';

// require - only sync modules are allowed
const theActualExportCommonJS = require('./common.js');
const theActualExportES6 = require('./es6.js'); // throws a helpful error message

// Both return promises!
require.import('./es6.mjs').then(exports => {});
require.import('./common.js').then(exports => {});
```

bmeck: would just wait on the `import()` fn personally: https://github.com/whatwg/loader/issues/149

002-es-modules.md
MylesBorins: /cc @nodejs/tsc I believe it is the date of the last update

002-es-modules.md
hybrist: 👍 Could this remove the need for overloading `require`? It just smells like the kind of issue that I'd have much rather fail then "succeed" ("huh? why is this export `undefined`? oh, because this was an ES6 module and I just got a promise"). Maybe that's my bias as someone who's pulled in to help people (often beginners) and this smells like yet another source of confusing "undefined is not a function" bugs.

bmeck: @jkrems if `import()` opens a path to load ESM from NCJS sure, until then need a way to load things / not gonna open up `require.import` as that is an additional migration if `import()` eventually lands.

002-es-modules.md
trevnorris: hm. I never defined that in the README. but sure. time of last update sounds good.

trevnorris:
@bmeck 

> `import` is url based

What about the case where one can `require('fs/')` to get a module named `'fs'` from `node_modules` instead of from the native set?

EDIT: nm. This looks to be covered with the `import 'abc/123';` example.

002-es-modules.md
```diff
@@ -0,0 +1,540 @@
...
+(https://tc39.github.io/ecma262/#sec-modulenamespacecreate)**
+    - Provides a means of creating a list of exports manually, used so that
+      NCJS `module.exports` can create `ModuleRecord`s that are
+      prepopulated.
+
+## 3. Semantics
+
+### 3.1. Determining if source is an ES Module
+
+A new file type will be used for ES Modules, `.mjs`. Files with this extensions
```
trevnorris: "Files with this extension" ?

bmeck: fixed

bmeck:
@trevnorris ah, need to make explicit what looks up things in core... are we trying to preserve being able to load non-core using that trick? several rules are already changing here, possible to block people doing such.

trevnorris:
@bmeck I'm not sure if doing `require('fs/')` is officially supported or not. When I'm on my computer, will look at existing tests.

002-es-modules.md
Jessidhia: As it is, this is just going to require people (and transpilers) to wrap every single `require` call in `Promise.resolve`, or at least `Promise.all`.

Updating modules to support ESM will also be a major breaking change, as it will change `require`s to return a completely different data type, and will require major refactoring of all consumers of a module, unless they all preemptively use `Promise.resolve(require)`.

``` js
import foo from 'commonjs/bar'
```

would have to transpile to:

``` js
Promise.all([require('commonjs/bar')])
.then(([foo]) => {})
```

even if no ESM imports are involved, as not doing so would prevent ever migrating `commonjs/bar` to ESM without breaking everyone.

---

tl;dr existing node modules that do just

``` js
const foo = require('bar')
```

will essentially lock `'bar'` into NCJS only if they don't want to break everyone come ESM support in node, unless they always wrap `require` in a `Promise` (and unwrap it to use).

bmeck: > As it is, this is just going to require people (and transpilers) to wrap every single require call in Promise.resolve, or at least Promise.all.

They do not need to and should not as that changes timing. ESM needs to internally unwind stack and return a `Promise`, changing every `require` to unwind the stack would be bad.

> Updating modules to support ESM will also be a major breaking change, as it will change requires to return a completely different data type, and will require major refactoring of all consumers of a module, unless they all preemptively use Promise.resolve(require).

This breakage also exists when people have to move to `import()` or `require.import()`

> Promise.all([require('commonjs/bar')])
> .then(([foo]) => {})

It needs to unwind prior to resolve, so wrap the arguments to `all`:

``` javascript
Promise.all([
  Promise.resolve().then(
    ()=>require('commonjs/bar')
  )])
.then(([foo]) => {})
```

All modules in the wild that Node can load are NCJS. It would be the babel/traceur/typescript/etc. output would change. No existing code would need to be changed. [Note if dependencies update and that changes their APIs, yes they need to changes are needed because a dep **changed**]

While I agree that not _always_ returning a `Promise<ModuleNamespace>` is a gotcha, there are other gotchas (including if `.mjs` would be loaded if a fallback was available in an API that doesn't load ESM).

I will patiently wait on that link I sent up thread. I don't see any advantage to `require.import()` if `import()` exists.

vanwagonet:
All this use of ESM for modules makes me think if there is a new filename extension, shouldn't it be `.esm`?

bmeck:
@thetalecrafter ES is the moniker for the specification of JS, but people use JS in conversation when not talking about the specification itself. Keeping the file extension similar to .js with the JS moniker was taken into account while making the decision on file extensions. Some considerations were listed in previous versions of this EP: https://github.com/nodejs/node-eps/blob/5dae5a537c2d56fbaf23aaf2ae9da15e74474021/002-es6-modules.md#511-reason-for-decision

martinheidegger:
@bmeck With https://github.com/bmeck/UnambiguousJavaScriptGrammar being accepted isn't the file-ending part obsolete? The problem now however is not that we can't identify what type a file is. It becomes a Ecosystem problem and I am not sure that is something Node/Core should be fixing.

bmeck:
@martinheidegger that was merged into the draft of the ESM proposal. The problem of determining [source text] type remains, and as stated in the current text of the proposal lack of acceptance by TC39 is in play here. The standard way to determine the contents of a file on any file system is to use a file extension. I have very little hope that my agenda item on TC39 will make any progress and am moving to what I see as the more realistic option.

bmeck:
I would also like to note existing talking points about differing from browsers being bad, accidental mode swaps (in part related to differing behavior) being bad, and security implications if there exists ambiguity.

matthewp:
Would you be open to a flag that enables "everything is an ESM mode"?

xjamundx:
@matthewp from my understanding currently there are issues with the spec that would mean for example: native code doesn't work at all with es6 modules, many core modules can't work at all with es6 mode. I may be way off here, but I think those are some of the reasons why that could not happen.

matthewp:
Ah, I wasn't thinking about core modules, ok then. :(

martinheidegger:
> I have very little hope that my agenda item on TC39 will make any progress and am moving to what I see as the more realistic option.

@bmeck I find it extremely sad to believe that. I have more hope tbh. Aside from that: If Node.js just assumes your proposal as accepted it might increase the chance for it to actually be accepted.

bmeck:
@martinheidegger Node can't move it to accepted unless the intent is to implement it that way, Node can implement the "spec extension" of requiring an import as noted in various places (https://twitter.com/awbjs/status/744700272190971904 , https://github.com/nodejs/node-eps/pull/33#issuecomment-229679824). However, this has implications that are problematic and Node very much would not like to have, including the ones listed above. If accepted I highly doubt the stance would change given there are already people stating the lack of need to change things.

martinheidegger:
I want [to publically state](https://twitter.com/leichtgewicht/status/773348056775266304) that there is a need to change things. It is an embarrassingly simple solution that would fix a major problem by creating a little effort for a minority of users and makes everyone's life better. Worth to champion for. _(I am not sure I said this enough: thanks for your efforts)_

Edit: Also posted it on [es-discuss](https://t.co/Gn3nDgXIUK)

002-es-modules.md
```diff
@@ -0,0 +1,540 @@
...
+The `.mjs` file extension will be searched for prior to any `.js` file in any
+algorithm that uses searching (`require` or `import`); e.g. once the Node
+resolution algorithm reaches file expansion, `path + '.mjs'` would be searched
+prior to `path + '.js'` when performing `require(path)`.
+
+# 3.1.1. Inter package loading using file extension breakage.
+
+There is knowledge of breakage for code that upgrades inner package
+dependencies such as `require('foo/bar.js')`. As `bar.js` may move to
+`bar.mjs`. Since `bar.js` is not the listed entry point this was considered
```
ljharb: Similarly, a common community "best practice" (although not a universal one) is to always omit extensions when requiring - to allow for "bar.js" to become "bar/index.js" transparently, for example - and require('foo/bar') would not break under this proposal.

HyeonuPark:
Desclaimer: I saw eps-002 so please read it before close it.

Not really sure this is the right place to write it, please tell me if not.

Currently [dynamic `import()` proposal](https://github.com/domenic/proposal-dynamic-import) landed stage 2 in tc39 process. This spec allows to use `import()` expression even within script context.

So my idea is simple: `import()` for es module, `require()` for others.

In this case, there's really no additional boilerplates are required, for both legacy codes and bleeding-edge codes. You can just `require()` to load scripts, jsons, and native modules as usual. You also can `import {foo} from 'module.js'` to statically load es module within another module code. And also, you can even use `import('module.js')` within script code as an expression which evaluated to promise.

That's it. No `.mjs`, no `"module": "index.js"`. I think this is the most elegant way to handle it.

I know this dynamic import proposal is not yet finalized. But as eps-002 is also stay in draft and not even started to be implement, I think we should discuss this before further implementation occurrs.

ljharb:
@HyeonuPark That would force asynchronous evaluation across the module boundary, and it would not allow existing unchanged code that has `require('foo')` to _continue working unchanged forever_ when the "foo" author refactors to provide ES modules. In addition, even with your suggestion the problem still remains of node needing to know whether a file is a module or a script, which brings us right back to `.mjs`.

HyeonuPark:
@ljharb the semantic of `require` should not be changed for that reason. 

If user types `require()`, traditional module loader is triggered, detects json/native module/es script via file extension, all operation unchanged.

If user types `import ... from ...` or `import()`, engine-native es module loader will parse given file as an es module.

And yes, users should update their codes when the library changed to the es module, just like when major version is updated. If they do not want to update their code, they can stuck in older versions. Of course libraries can support legacy modules via `require('libname/legacy')` like name.

ljharb:
That's a dealbreaker. Old code must not need updating when a module switches module formats.

HyeonuPark:
IMHO, updating library from script to module should be considered as breaking change(semver major version update) as it changes how user consume the library. This literally means api change - not only in semantically, but also in syntactic form.

If you don't agree with the statement above, please answer why changing syntax should be treated as non-breaking change.

ljharb:
Because the syntax is an implementation detail that the consumer need not care about in the majority of cases. If I change from a function expression to a function declaration, but either way export the same function, then that change is not semver-major - it's unobservable. If upgrading to ES modules requires a breaking change, then a) many people will never do it, and b) the ecosystem will be forked between "modules that require a new node or browser" and "modules that work everywhere" - we don't want a Python 3 situation here.

hybrist:
@ljharb There's no Python 3 situation unless there's _no_ interop. Switching to ES modules will drop support for a bunch of browsers, many versions of node, older (all?) versions of browserify... It's almost necessarily a breaking change for every library.

And it's not like "switching to the ES6-equivalent has observable consequences" is without precedence. ES6 classes behave differently than the usual, hand-written ES5 equivalent. Differences that can be observed by users of the class. And switching to class syntax almost always requires a major version bump (for the compatibility reasons mentioned above). And every library depending on your code will _also_ do a major version bump.

None of that means there's a Python 3 situation.

ljharb:
@jkrems fair counterpoint. but every breaking change is added friction that interferes with adoption. Also, switching to class syntax only requires a major bump if you weren't properly using inheritance in ES5 (which is admittedly easy to do wrong) - you can also make the major bump by fixing your ES5, and then have a patch bump to switch to class syntax. Without the interop I'm talking about, there's no way to avoid the breaking change.

hybrist:
Sorry, I should've been more explicit in my classes example: Even if you were doing ES5 class inheritance perfectly, you might have relied on things like enumerating methods on the prototype chain. And suddenly all methods "disappear" because a library switched to ES6 classes. Or you were expecting that the `__proto__` of a class is always `Function.prototype` (bad code - but possible to write). Thankfully ES6 made some improvements to default semantics of JavaScript, I don't think those should be treated as annoyance that can or should be "optimized" away.

The reason why I objected to the Python 3 comparison is that it effectively turns this discussion into "my solution or you want to break the ecosystem". Which I don't think is fair. There's a tradeoff here between various degrees (and kinds) of pain and getting the most out of the changes that ES6 brings.

I'm honestly concerned about the "let's just magically switch the implementation of a library based on the version of node you're on, deep in the dependency tree". _That_ to me sounds like a potentially ecosystem-breaking change on the day that new version of node rolls out. Because there's no way to predict what it will break (because that code never ran before). And at the same time it's not clear to me what benefits it would still have. If your ES6 module is "proxied" by a CommonJS one, no bundler can properly tree-shake it (not anymore than local dead code elimination could anyhow). So yes, you're running that code in node and it behaves exactly like before. But the only thing you gained is a bit of syntactic sugar. Which, to me, doesn't sound all that compelling given the risk implicit magic adds.

paulwalker:
Can someone please tell me what the hold-up is now w/ ES6 module support? It seems as if there is argument about "breaking changes" just for "syntactic sugar," and that this diminishes the real desire of developers to move things forward without transpilation. Is it not a matter of when and how node supports ESM, correct?

bmeck:
@paulwalker you might want to jump towards the end of the broadcast for https://github.com/nodejs/CTC/issues/59 . There are breaking changes that would require specification changes to JS to support the currently accepted draft. That draft is a synchronous loader that supports named imports from CJS. This PR is a secondary option should those things which act like babel no longer be viable. There were 2 main holdups until last CTC meeting, now there is 1.

* As of a couple of weeks ago, named imports from CJS are off the table for now. It would require property delegation on variable access which has performance, and implementation problems that affect things not even using CJS. Developers will need to change code from `import {Component} from 'react'; ... Component ...` to `import React from 'react'; ... React.Component ...`.

* Synchronous / asynchronous is mentioned in CTC meeting as needing to be discussed now. With one of the main reasons to match the majority of transpilers being removed from discussion the question of wether to match the async nature of the web becomes the last item to talk about.

bmeck:
ah, to note: for a synchronous loader to be possible https://github.com/caridy/proposal-dynamic-modules needs to land in JS spec.

WebReflection:
I am not sure if I should file a new EP document or add [my proposal](https://github.com/nodejs/node/issues/10956) to this thread but since this is very similar to what I've done I'll give it a try.

### module.import(path):Promise
It seems that everyone agreed at least in two points: the `import` name, and the `Promise` as "protocol".

Since it's impossible to polyfill a reserved word like `import` is, my solution is to bring the `import` method to the `module` object, and for the following reasons:

  * `require.import()`, as action, does not sound as natural as `module.import()`
  * the module is where we `exports`, having a single responsible object to both export and import seems reasonable, specially for code that won't ever need to even `require` another module, but it'll surely need to import a module and/or export one

#### Implementation
Accordingly, the following 3 LOC are theoretically all it takes to enable asynchronous module import, as well as asynchronous module exports:
```js
// Module as module.constructor
Module.prototype.import = function (path) {
  return new Promise(res => res(this.require(path)));
};
```

Considering a generic, good old fashioned, NodeJS module like the following one:
```js
// modules as we know: converter.js
const crypto = require('crypto');
module.exports = {
  sha256: (str, secret = '') =>
    crypto.createHmac('sha256', secret)
          .update(str)
          .digest('hex')
};
```

It'll be completely transparent for developers to either include it synchronously:
```js
// same good old synchronous flavor
const converter = require('./converter');
console.log(converter.sha256('yolo'));
```

or asynchronously:
```js
// module.import asynchronous flavor
module.import('./converter')
.then((converter) => {
  console.log(converter.sha256('yolo'));
});
```

### Asynchronous exports
The pattern automatically enables the ability to also export module asynchronously, simply returning a Promise as exports.
```js
// generic async export
module.exports = new Promise(define => {
  // some asynchronous operation ...
  define({my:'module'});
});

// asynchronous converter.js
module.exports = module.import('./crypto')
.then(crypto => {
  // return the module to export
  return {
    sha256:(str, secret = '') =>
      crypto.createHmac('sha256', secret)
            .update(str)
            .digest('hex')
  };
});
```

As previously mentioned in this thread too, it's straight forward to require multiple modules too:
```js
module.exports = Promise.all([
  './awe',
  './some'
].map(
  m => module.import(m)
)).then(modules => {
  const [awe, some] = modules;
  return {awe, some};
});
```

Last, but not least, I've already implemented a [tiny utility](https://github.com/WebReflection/common-js#commonjs--moduleimport-) that brings this ability to both NodeJS and browsers. You can use it as playground also importing by default from [unpkg CDN](https://unpkg.com/) whenever a module is imported, or required, without relative or absolute path.

Last part, is the only one I'd like to better think about, since current implementation works well but it's coupled with unpkg service, but I believe there should be a better mechanism to intercept upfront non relative paths and eventually map them to specific files.

I hope any of this can help moving forward with asynchronous module loader, this could also be the best way to polyfill, whenever TC39 will finalise it, the global `import` through transpilers.

Best Regards

bmeck:
@WebReflection I would make a separate PR since that is not related to supporting ESM itself. It sounds like it just wants a standard polyfill for `import()`, doesn't appear to have any different features? To discuss:

* The auto awaiting of module.exports was something brought up sometime in the past year, but never fully researched to see if it affects existing modules.
* Will need to explain why this polyfill needs to be in core rather than runtimes of transpilers.
* Might want to use the name `require.import()` as historical `require.async()` was a thing that tried something similar in the early days of node.

WebReflection:
Will open a new PR then. To quickly answer your questions:

  * not sure I understand the first point but there's no auto awaiting of anything here, it works already as it is
  * transpilers are not always an option, having this in core means reaching every CommonJS user
  * the `require` function has been [abused](https://www.npmjs.com/package/require.async) and [polluted](http://requirejs.org/docs/api.html#data-main) for the last _N_ years to accomplish something similar but never `Promise` based. Moreover, semantically speaking, you import a module, rather than requiring an import (of a module).

Will try to file the EP ASAP, thank you.

002-es-modules.md
```diff
@@ -0,0 +1,547 @@
...
+
+1. Allow a common module syntax for Browser and Server.
+2. Allow a common registry for inspection by Browser and Server
+   environments/tools.
+    * These will most likely be represented by metaproperties like
+      `import.context`, but the spec is not yet fully in place.
+
+## 2. Related
+
+[ECMA262](tc39.github.io/ecma262/) discusses the syntax and semantics of
```
targos: `https://tc39.github.io/ecma262/` (or the link will be relative)

002-es-modules.md
```diff
@@ -0,0 +1,547 @@
...
+JavaScript with IANA as an official file type needs to have this extensions
+added. This comes from browsers being unable to implement a new MIME for the
+differing parse goals. In order to make this change, contact the IESG.
+
+The `.mjs` file extension will be searched for prior to any `.js` file in any
+algorithm that uses searching (`require` or `import`); e.g. once the Node
+resolution algorithm reaches file expansion, `path + '.mjs'` would be searched
+prior to `path + '.js'` when performing `require(path)`.
+
+# 3.1.1. Inter package loading using file extension breakage.
```
targos: `####`

002-es-modules.md
```diff
@@ -0,0 +1,547 @@
...
+  * for CLI usage, it is the absolute URL for the current working directory
+
+* if `request` parses via the [URL parser](https://url.spec.whatwg.org/#concept-url-parser)
+  * let `url` be the result
+* else if `^[.]?[.]?[/]` prefixes `request`
+  * let `url` be the result of `new URL(request, context_url)`
+  * if `url` points to a directory
+    * let `url` be the result of searching the directory
+  * if `url` does not point to a file
+   * for the well known file extensions `extension` in `[
```
targos: Indentation is off. The next lines will also need to be adjusted

002-es-modules.md
```diff
@@ -0,0 +1,547 @@
...
+If `require` resolves to an ESM source text, the following steps are taken.
+
+1. let `ret` be a newly created Promise
+2. queue a load job on the event loop
+  1. if loading completes normally resolve `ret` to the ESM Module Namespace
+  2. if an error occurs reject `ret`
+3. put a "fetching" entry in the ESM cache
+4. return `ret` 
+
+##### 5.4.1.1. Examples
```
targos: `3.5.3`?
Next titles have to be changed accordingly

002-es-modules.md
```diff
@@ -0,0 +1,547 @@
...
+### 5.6. Known Gotchas
+
+All of these gotchas relate to opt-in semantics and the fact that NCJS has a
+dynamic loader while ESM has a static loader.
+
+No existing code will be affected.
+
+#### 5.6.1. ESM exports are read only
+
+The objects create by an ESM are [ModuleNamespace Objects][5].
```
targos: nit: created.
Is `[ModuleNamespace Objects][5]` supposed to be a link?

targos (dismissed):

targos:
@nodejs/ctc Should we vote here about the asynchronicity of the ESM module loader?
My vote is +1.

addaleax:
I’m +1 on an asynchronous loader (*sigh*).

misterdjules:
My vote is: abstain.

MylesBorins:
+1 for async loader for `import`.

Primary motivation is to avoid behavior differences between node + browser 

Is it reasonable to keep the door open for a sync loader in compatibility apis such as `require('esm')` or `require.import('esm')` In these instances we are not dealing with potential behavior differences, especially if the sync operation is using the async import implementation under the hood.

evanlucas:
I'm -1 for async loader for import. IMO, large differences between node and the browser already exist and I feel that an async loader makes the transition from ESM syntax with babel more difficult. I think we should stick with a sync loader.

Trott:
Vote count for async vs sync loader. (I'll try to keep this comment up-to-date)

Async: @targos @addaleax @MylesBorins @ofrobots

Sync: @evanlucas 

Abstain: @misterdjules @Trott @mscdex @rvagg @bnoordhuis @Fishrock123 @thefourtheye 

Minimum number of additional CTC votes needed to come to resolution: 3
CTC members who have not yet voted: @ChALkeR @chrisdickinson @cjihrig @indutny @jasnell @mhdawson @shigeki @trevnorris 

(Probably doesn't make sense to add a vote count at the top because it will look like a vote count for the EP and not for the async vs. sync issue specifically. So putting it here.)

mscdex:
My vote: abstain

ljharb:
(non-CTC/TSC member here) I don't actually care if the loader is sync or async - async seems nice in that it maintains more parity with browsers. What I do think is very critical is that `require('foo')` work the same whether `foo` is CJS, or ESM - iow, I want "migrating to ESM" to be as seamless as possible, and not a breaking change. If this (ie, `require('esm')` not returning a promise, but just returning the default export or a Module Record)  can be achieved with an async loader, then I think that's clearly the far superior approach (also, assuming that top-level `await` requires an async loader, which isn't something I grok yet).

@bmeck is there a possibility that async loading could allow `require('esm')` to return the default export or Module Record? Can you explain why top-level `await` is impossible with a sync loader?

ofrobots:
My vote for 'should the module be sync or async?': 'async'.

Although we are not debating this issue yet, IMO `require('esm')` should throw. There is expectation of sync behaviour with `require`. A dynamic `import()` of esm from cjs would be fine by me.

ljharb:
@ofrobots if it throws, then the net effect will be that nobody ever refactors their modules to use ESM, and ESM will be DOA.

rvagg:
Abstain from me, I've had @Fishrock123 explain it in more detail to me but I still can't form an opinion because it seems the tradeoffs are too large either way.

Fishrock123:
> if it throws, then the net effect will be that nobody ever refactors their modules to use ESM, and ESM will be DOA.

I do not think that the behavior of `require()` under an async loader matters at all. If anything, throwing would be better because it is explicit and we can tell the user what to do to use ESM.

ljharb:
@Fishrock123 why would any module author (such as myself) want to migrate my module to ESM if it means that users using n ESM-supporting node wouldn't be able to use it with `require`? (ie, they'd be forced to refactor to ESM in order to consume my package)

I'll simply never refactor to ESM, and everything will continue to work perfectly, and that's what the majority of authors will likely do as well (since they tend to care about their users' ability to use their packages).

Fishrock123:
@ljharb If we implement an async loader you will have to refactor regardless.

So what is better then? Two possible behaviors of `require()` depending on what is in the file you are asking for? Or two APIs for for each type of module with explicit throws?

Fishrock123:
I'm going to abstain. All the options are pretty awful IMO and I don't really want to be responsible for signing off on any of this really.

ljharb:
I don't know what's the best path - I'm just convinced that if *consumers* have to know or care about what module format I'm using, it's going to be a disaster for everyone. I'd rather node never ship ESM, than fork the ecosystem by forcing users to handle my chosen module format.

Fishrock123:
> I'd rather node never ship ESM, than fork the ecosystem by forcing users to handle my chosen module format.

This is part of the reason why I had/have been a proponent of a Sync loader for so long...

Keep in mind you need to chose regardless if you are using an ESM with an Async loader because you will need to convert your entire load order to async.

Trott:
No official vote resolution on the async vs sync loader, but the sense of the group at the meeting today is that CTC members (or at least the ones who feel informed enough to have an opinion) are horribly conflicted but are mostly leaning towards **async**. We would like to tentatively move that way to at least get a better idea of what an implementation would look like.

WebReflection:
FWIW I am discussing an always asynchronous `module.import('anything')` that would solve half of the issues raised in here as opt-in, and I also agree `require` should be kept synchronous.

Just my 2 cents

joshgav:
Initial implementation discussion continues in https://github.com/nodejs/node/issues/11233.

flying-sheep:
`.mjs` is a bad idea. The standardized version should be `.js`.

Counter-proposal: ES6 modules are a separate world from commonjs modules:

Bridges are explicit and synchronous: `import {require} from 'commonjs'` and `module.import('some-es6-module')`.

When you want to profit from ES6 static guarantees, use and write code that has them.

Everything is still `.js`, clear error messages tell you that you just tried to use the wrong import mechanism.

ljharb:
@flying-sheep the sole purpose of a file extension is to define how to parse a file. Modules are a new parse goal, thus they need a new file extension - otherwise why have extensions anywhere?

WebReflection:
@ljharb sure, but in this case the extension doesn't define the file purpose: a JavaScript module.

I just wonder: if the direction is to load new modules asynchronously, what's the real issue in taking "_all the time you need_" to pre-fetch the target file and statically understand if that's using ES6 import / exports or the `module` global in it?

I fully understand this is nothing suitable for `require` but ... who cares? Why cannot CommonJS implement a promise based `module.import(path):Promise` that once the top async behavior will be in place would make things as simple as it is today for require?

In that case, the import is resolved only once the module kind has been defined, and the right mechanism to import it applied.

After all, we know the extension has never been a good guard for file types, so that everything needs to be fetched first, analyzed in terms of meta bytes, and eventually rendered accordingly with that analysis, not just the extension, isn't it?

I mean, a .png that is actually a .gif or vice versa will be rendered correctly from a browser anyway so ... the extension is historically a not fully trustable hint.

We can do better, since we've decided the approach is async.. No need to rush conclusions based on file names and extensions, IMHO.

unional:
Agree. And on the browser side there are ways to do this like `systemjs` which follows WHATWG.

flying-sheep:
extensions are just hints for the mime type. mime types aren’t generally distinguished between “module” and “script”. if this were the case, why not use `.cjs` for all commonjs modules, which need `module`, `exports`, and `require`? or `.bjs` for scripts that have access to all the browser APIs?

as said: my main point is that i don’t think commonjs and ES6 modules should interoperate using the same syntax. and once “import commonjs module from ES6” and “import ES6 module from commonjs” aren’t using the same syntax(es) as “import ES6 module from ES6” or “import commonjs module from commonjs”, we don’t need to invent a new file extension.

shipping both variants could be possible via `package.json`; `some-hybrid-module`’s `package.json` contains `"main": "bundle-commonjs.js", "module": "bundle-es6.js"`

`common.js`:
```js
const es6 = await module.import('some-es6-module')
const hybrid = require('some-hybrid-module')
const commonjs = require('some-commonjs-module')
```

`es6.js`:
```js
import {require} from 'commonjs'
import hybrid from 'some-hybrid-module'
const commonjs = require('some-commonjs-module')
```

WebReflection:
this:
```js
const es6 = await module.import('some-es6-module')
```
is exactly what I've [proposed](https://github.com/nodejs/node-eps/pull/50) but I've found a lot of skeptical/refractory opinions while to me is the easiest possible way to move forward and it plays well as migration pattern.

flying-sheep:
my proposal is that this function only works with ES6 modules, though, in order to separate the ES6 and commonjs module worlds.

WebReflection:
like i've said, the moment you load asynchronously something is the moment you can also statically analyze what is it (a bit like reading mime type information).

A parser that drops strings and comments and see if there is an `/^export /m`in the code doesn't seem like such expensive operation to perform, or is it? If possible, I don't see why we should have two ways to load CJS or ES6 modules through `module.import`

hybrist:
@WebReflection That approach (analyze the module source to determine the parse goal) has been discussed a lot in the past. The problem is that there are scripts that are ambiguous: scripts that are valid CommonJS & valid JS modules but would do different things when executed (e.g. `Object.freeze({}).x = 42;`).

WebReflection:
> there are scripts that are ambiguous

this looks like a developer issue, not a `module.import` one. If there is a single `export ` is ES6, otherwise it's CommonJS, I don't see any ambiguity but if already discussed I'm sure I'm missing something and I'd like to read more about it.

Hopefully transpilers were out of the equations since these should solve problems, not create new.

hybrist:
There is - currently - nothing in the ES module spec that would force ES modules to import or export anything. The example I gave above was a complete module. It's possible to have a module that consists only of global side effects. And such a module would either run in strict mode (if interpreted as ES module) or not (if interpreted as CJS module). It's impossible to tell what the *intended* runtime behavior was, just from looking at the source code. It's for the same reason that you wouldn't (shouldn't) run node with the `--use-strict` flag.

WebReflection:
I give for granted any code I run is strict, and if the idea is to fix the gap between ES6 and CJS modules with a new `module.import` API proposal, I wouldn't mind at all, and I would actually expect, whatever I'm importing will run in strict mode.

This wouldn't play 100% well in terms of backward compatibility, but I really would like to know numbers behind the "don't want to run in strict mode" in CJS panorama, specially considering that is always possible to opt in no-strict mode via `Function`, when/if really needed.

So I still believe it's a developer issue. There could be some caveat and compromise but I see 99% of use cases happily working all together and moving forward VS something ugly as a new `.mjs` extension would be.

At this point I'd rather say next version of node always load `.cjs` modules as CJS and all new modules as ES6, that'd be more sensible long term solution for the future since nowadays most developers transpile to target older engines, changing dest file name would be quite simple (yet I think `module.import(anything):Promise` is a better solution)

flying-sheep:
my proposal was that `module.import` only loads ES6 modules (and gives a special, helpful error message when you try to load a commonjs module). and ES6 modules are always strict mode.

loading commonjs modules asynchronously is useless, so `require` is fine for those.

WebReflection:
> loading commonjs modules asynchronously is useless

It's about having a single entry point to load modules, no matter how these have been published.
How useless is that?

Moreover, asynchronous import enables asynchronous export too.
```js
export default await Promise.resolve().then(doStuff);
```
This can be transpiled into `module.exports = Promise.resolve().then(doStuff)`

How useless is that?

michael-ciniawsky:
```
| – lib
|   | – index.js
|   | – module.js // ESM
|   | – legacy
|   |   | – index.js
|   |   | – module.js // CJS
| – package.json
```

**package.json**
```json
{
  "name": "package",
  "version": "1.0.0",
  "main": "lib/legacy",
  "module": "lib"
}
```

**lib/index.js**
```js
import module from './module'

export default {
   module
}
```

**lib/legacy/index.js**
```js
const module = require('./module')

module.exports = {
  module
}
```

No interop between CJS and ES2015 `require('lib/legacy/*')` or `import * from 'lib/*'` 

**CLI**
```bash
node require.js
node --module module.js
```

**Server**
```js
// ES2015
import modern from 'lib'
// CJS
const legacy = require('lib/legacy')

// ES2015
const module = import('lib/module')

module.then((module) => module())

// CJS
const module = require('lib/legacy/module')

module()
```

**Browser**
```html
<script src="lib/legacy/index.js"></script>
<script type="module" src="lib/index.js"></script>
```

WebReflection:
@michael-ciniawsky that is still trying to compare potatoes and tomatoes without considering that asynchronous import means also asynchronous export when you have asynchronous dependencies in your module.

This might be considered a non issue on node-land but it's impossible to load synchronously modules on the browser due deprecated synchronous xhr on the man thread.

This direction will force everyone to bundle and it will penalise a lot the Web.

michael-ciniawsky:
@WebReflection Could explain this a bit further please ? :)

WebReflection:
@michael-ciniawsky how are developers going to export modules that use dynamic `import` which is asynchronous?

```js
export default await Promise.all([
  import(mod1),  // parallel asynchronous imports
  import(mod2),  // as opposite to N implicit await
  import(mod3)   // ideal for HTTP2 scenarios
]).then([mod1, mod2, mod3] => {
  const myModule = {
    doStuff() {
      return mod1() + mod2() + mod3();
    }
  };
  return myModule;
});
```

There's no way you can use `require` to import a module like that, you need something else that better scale, like the `module.import` I've discussed already.

If we keep transpiling ES.next modules into `require` we'll fail everyone expectations and intent to ship asynchronous modules on the Web.

I know pure NodeJS devs don't care, but npm is used every day more as universal repository, not only server side and its synchronous, quite outdated, `require` mechanism.

I hope I've better explained what I meant.

michael-ciniawsky:
**lib/*.js**
```js
exports default function module1 () => {}
```
```js
exports default function module2 () => {}
```
```js
exports default function module3 () => {}
```

**browser/static.js**
```js
import module1 from 'lib/module1'
import module2 from 'lib/module2'
import module3 from 'lib/module3'

const use = fetch(url).then((json) => module1(module2(module3(json))))
```

**browser/dynamic.js**
```js
const use = fetch(url).then((json) => {
   return Promise.all([ 
     import('module1'), 
     import('module2'), 
     import('module3')
   ]).then((...modules) => module1(module2(module3(json)))
})
```

Why export an asynchronous module ? Do I miss something here ? Isn't the dynamic `import()` proposal supposed to return a Promise anyways/always ?

```js
export default await Promise.all([...modules]) // Is this valid?
```
```js
export default async () => await Promise.all([...modules]) // ?
```

> If we keep transpiling ES.next modules into require we'll fail everyone expectations and intent to ship asynchronous modules on the Web.

There should be no transpiling evolved at all, if code is transpiled to CJS you `require()` it and require is synchronous ¯\_(ツ)_/¯. Static/Dynamic import (should work) works only with ES2015 Modules. Also there is no native support for CJS in browsers

> I know pure NodeJS devs don't care, but npm is used every day more as universal repository, not only server side and its synchronous, quite outdated, require mechanism.

I'm not a 'pure' NodeJS Dev 😛 , I joined the discussion here, bc I'm worried about the `.mjs` proposal which would be incompatible with ES2015 modules `.js` in the browser  **forever**.
There should be a migration path for node achieved with `pkg.module > pkg.main` over time and as you mentioned npm packages, if not requiring node core modules like `fs`, should be consumable in the exact same way in node && browsers without any setup.

WebReflection:
> Why export an asynchronous module ? Do I miss something here ? Isn't the dynamic import() proposal supposed to return a Promise anyways/always ?

Exactly, and if you export a Promise you have an asynchronous export: it's an implicit consequence.

The following:
```js
import module1 from 'lib/module1'
import module2 from 'lib/module2'
import module3 from 'lib/module3'
```
Is fundamentally different from the following:
```js
Promise.all([ 
     import('module1'), 
     import('module2'), 
     import('module3')
])
```

> Is this valid?

It was an example, the point is in exporting something asynchronous that cannot be transpiled into a `require()`.

> require is synchronous ¯\\_(ツ)_/¯

hence not a suitable transpilation target for asychronous modules.

> Static/Dynamic import (should work) works only with ES2015 Modules.

No, why is that? It can be transpiled keeping asynchronous semantics/nature.

> Also there is no native support for CJS in browsers

It never mattered, did it? Browserify works already for `require`?

Today we have bundlers, tomorrow we could use [this utility](https://github.com/WebReflection/common-js#commonjs--moduleimport-) which is already asynchronous and it already solves the issue and it brings CJS to browser, or anything similar, really, brought in by bundlers or transpiled.

>  I'm worried about the .mjs proposal 

so am I

> There should be a migration path

which is what I've explained already via `module.import`

michael-ciniawsky:
> Is fundamentally different 

Yep, I understand, but isn't that handled by `import()` ?

> It was an example, the point is in exporting something asynchronous that cannot be transpiled into a require()

Why would somebody want to do that? :) When node supports `import()`  => use `import()`. `require()` is 'legacy' 😛.

> hence not a suitable transpilation target for asychronous modules

Yep, so may proposal here is to *not* even try :D

> It never mattered, did it? Browserify works already for require?

Because Browserify, webpack, Rollup transforms CJS to e.g an IIFE, if you have a CJS codebase => keep using one of this tools. Otherwise enjoy `<script type="module"></script>` and forget about them in this context

> which is what I've explained already via module.import

BTW I'm not against `module.import` in particular, so no offense here, but if I hear words like `transpile`, I'm skeptical by default 🙃

> so am I

👍 💯

WebReflection:
We're not adding much so I'll stop here, but:

> if I hear words like transpile, I'm skeptical by default

Your proposal will be brought in by tools and tools that will transpile and generate the legacy part 'cause I don't think anyone will ever write that hierarchy by hand with duplicated modules.

Hence my underlying that `require` is not a good target for asynchronous modules, we need a better mechanism in CommonJS land.

I honestly don't understand what's so difficult to digest about it (not talking about you, but in general)

michael-ciniawsky:
> Your proposal will be brought in by tools and tools that will transpile and generate the legacy part 'cause I don't think anyone will ever write that hierarchy by hand with duplicated modules.

Well, I can't really tell (that would be speculative), maybe you're right, as long as `.mjs` is not going to happen / can be avoided anything better is highly welcome and appreciated I guess. My repetition of the `package.json` proposal by Yehuda Katz && Dave Herman should only function as a reminder :D, it's by no means the only solution, but the most promising imho.

flying-sheep:
OK, so let me rehash my idea with new information.

Goals
------

1. CommonJS modules need to work without change (no `.cjs`)
2. ES6 modules need to work without change (no `.mjs`)

Two worlds and explicit bridges
-----------------------------------

Module type | Exports | Imports
------|------|------
CommonJS | `module.exports = CJS` | `require('CJS')`<br>`await import('ES6')`<br>`module.import('ES6')`
ECMAScript 6 | `export [default] ES6` | `import [... from] 'ES6'`<br>`await import('ES6')`<br>`import require from 'commonjs'; require('CJS')`

`require()` can’t import ES6 modules. (NodeJS does not get top-level `await`, so we add a synchronous `module.import()` that serves that role). `require('mod')` will result in a syntax error if `mod.js` contains `export` or `import` statements.

`import [... from]` and `import()` can’t import CJS modules. (they don’t benefit from asynchronous loading anyway). `import 'mod'` will result in a “name undefined error” if `mod.js` does not define `module` or `require`, but uses them (e.g. if it’s accidentally a CJS module someone tries to import with a ES6 mechanism).

Hybrid modules and packages
-------------------------------

if a file `mod.js` exists for its side effects and contains no exports or imports of any kind (or only `import()`), it’s both a valid ES6 and CJS module and can be imported by both. (It could behave differently due to implicit strict mode in ES6 modules, but there’s no way around it: The preceding sentence is no proposal, but already fact)

A npm package’s `package.json` can contain the `main` and `module` entry points for CJS and ES6, respectively, turning *the package* hybrid (but not its modules! `require('pkg')` will import a different file than `import 'pkg'`!)

builtin node packages like `fs` will behave that way and be importable with every mechanism.

Porting
--------

A consequence is that modules written in ES6 but relying on babel or so to transform `import` into `require()` (transforming them into CJS modules) can be converted into true ES6 modules.

For this, the build step will be removed, and each imported non-hybrid package has to be replaced manually or by `package.json`-aware codemod: `import foo from 'cjs-only'` → `import require from 'commonjs'; const foo = require('cjs-only')`

Disadvantages
---------------
* `module.import` (or the same thing under another name` needs to be added
* the pitfall of importing side-effect-only modules with (`import`) or without (`require`) strict mode

bmeck:
@flying-sheep 

> NodeJS gets top-level await, so import() serves that role

note that while `import()` is coming to Script/CJS top level await can't since it changes run to completion. It would mean CJS would not run to completion in a blocking manner. With Node version 7.6 it is possible to change the `.js` file wrapper to an async function to achieve the proper grammar, but it breaks how CJS works fundamentally since all CJS would return Promises.

> import [... from] and import() can’t import CJS modules

I assume that builtins could still be loaded even though they will stay CJS.

> import 'mod' will result in a “name undefined error” if mod.js accesses module or require.

Unclear what this means, you mean if they use variables named `module` or `require`? What would `var require = () => {};` in your source text mean wrt this error?

> if mod.js exists for side effects and contains no exports or imports of any kind (or only import()), it’s both a valid ES6 and CJS module and can be imported by both.

Note, this can change the behavior of a source text.

flying-sheep:
> note that while `import()` is coming to Script/CJS top level await can't since it changes run to completion

damn, makes sense. i’ll add `module.import` again. /edit: done

> I assume that builtins could still be loaded even though they will stay CJS.

i assume so too. i added a bit about hybrid *packages*, which would fit here.

> Unclear what this means, you mean if they use variables named `module` or `require`?

i meant that if you accidentally use a ES6 import mechanism to import a CJS file, it will just encounter access to undefined variables names `module` or `require` and throw the usual error. defining them of course works normally.

> Note, this can change the behavior of a source text.

yes, you need to be aware of this pitfall (or `use strict` in all CJS), but there’s no way to avoid it AFAIK. files without imports/exports *are* valid CJS and ES6 modules already.

/edit: clarified everything

ljharb:
"require can't import ES modules" makes it a nonstarter imo, because that means I can't transparently refactor my CJS module to an ES module, which is a critical goal.

bmeck:
@ljharb that can be ok as long as poly packaging works so you can ship a CJS form as well.

flying-sheep:
@ljharb that’s not true.

1. ATM all packages are CJS or hybrid, so convert all `require('relative-path')` calls to `import ...`
2. once there are ES6-only packages you need to know what they are when porting, no matter which proposal gets implemented.

bmeck:
@flying-sheep the concern is for people consuming your package. If you are a module author of `foo`. You need people to continue to be able to use `require('foo')` even if you also support `import('foo')` so you don't break people consuming your module when you upgrade to ESM.

flying-sheep:
yeah, as you said: that’s easy due to hybrid packages.

bmeck:
@flying-sheep in order to support things like deep linking ala `require("lodash/chunk")` you need to have the full set of things just like [Defense of .js](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md)

flying-sheep:
yeah, i always hated that `main` doesn’t work as root for paths like this.

michael-ciniawsky:
@bmeck Will something like `module.import()` or `require.import()` be implemented ? 😕 I think they are adding even more noise and confusion. Nice work so far, the proposal is becoming much cleaner, still emphasizing to reconsider `.mjs` if possible 😛 

> as long as poly packaging works 

? Is `package.json` not 100% of the table, or I'm not getting it 🙃 ?

bmeck:
@michael-ciniawsky any solution that encapsulates use cases and is not prohibitive to adoption or education is open for discussion. Defense of .js is a very well thought out solution, but remains complex in order to achieve high use case coverage.

As per `*.import()` I don't think it will be shipped in light of `import()` coming down the pipe.

[edit] note that a goal is hopefully to allow a path that *all* new code can be written in ESM without causing a perpetual burden to either Node or Browsers.

michael-ciniawsky:
>As per *.import() I don't think it will be shipped in light of import() coming down the pipe.

👍 

What parts of Defense for '.js' are prohibitive ?

> but remains complex in order to achieve high use case coverage

?

bmeck:
> What parts of Defense for '.js' are prohibitive ?

* Not all packages have "main" (somewhat rare)
* Hesitancy towards `package.json` changing how files are interpreted. Means checking `package.json` becomes something to do if you ever are in doubt. If code uses `import()` the intention is still unknown.
  * part of hesitancy is that not all CJS can upgrade
* `"module"` swaps the whole package unless you have a `"main"`. kind of odd if you add a `"main"` later on to support CJS.
   * [Fallthrough behavior of require](https://github.com/bmeck/node-require-fallthrough-example/blob/master/nested/index.js) gets confusing is swapping modes as you fall through.
* `"module.root"` changes pathing behavior in odd ways sometimes, like how `import("../")` wouldn't escape your package it would escape the "module.root"
  * this seems a bit confusing as it does add a pathing manipulation, confusion when reading `import "foo/bar"` and it not mapping to disk.
  * doesn't allow easy mixing in the same directory (if porting a large app you have to place all ESM in the root and all CJS outside the root)
  * an interesting and useful feature even w/o "module", question of if it is in scope
* File based CLI usage without package.json
* Learning behaviors
  * differing way to determine/express how to run things w/ node and browser that are not readily visible (must learn both)

As has been stated many times, the decisions here are about weighing cost/advantage of all the things. There are also downsides to `.mjs` that have been discussed as well. The prevailing thoughts have been that it is the simple and does not leave lasting persistent burden when moving between environments or learning for the first time.

unional:
> "module" swaps the whole package unless you have a "main". kind of odd if you add a "main" later on to support CJS

IMO as things moving forward, this scenario would be rare. 🌷

michael-ciniawsky:
> "module" swaps the whole package unless you have a "main". kind of odd if you add a "main" later on to support CJS. 

That shouldn't be the case of course, `module` applies where it is set. Everything else is treaten like `main` (CJS)

> "module.root" changes pathing behavior in odd ways sometimes, like how import("../") wouldn't escape your package it would escape the "module.root" 

`module.root` is awkward. Would it be mandantorily needed? e.g => `module: [ entry.js, lib ]` etc.

> doesn't allow easy mixing in the same directory (if porting a large app you have to place all ESM in the root and all CJS outside the root

kk, is separation not even better ? Why in `root` only, is in different directories not enough?

> File based CLI usage without package.json

`node --module module.js`

**$HOME/.noderc**

`node module.js`
 
```
{
   "module": true // node => node --module
}
```

bmeck:
@michael-ciniawsky 

> module.root is awkward. Would it be mandantorily needed? e.g => module: [ entry.js, lib ] etc.

There were discussions of using `glob` like patterns to whitelist which files should be treated as ESM. This is the `"modules"` field I didn't bring up in last comment.

> $HOME/.noderc

Would be adding a file in a manner that we haven't seen before. Unsure how core would feel about making a settings file standard. You could bring it up though, might be useful even w/o ESM.

> kk, is separation not even better ? Why in root only, is in different directories not enough?

This is complicated, it depends on what you setup in `"modules"` if you have `"main"`. Reading the proposal would explain this.

> File based CLI usage without package.json

`--module` is also needed for `.mjs` for STDIN/`-e`/`-p`, but `.js` for package.json based approaches do not give information on how files are supposed to run. Reading the file to figure out if it is ESM or not is necessary, even then if ambiguous, check the docs.

On package.json: we add 3 new fields in package.json (which are interdependent), potentially a settings file, and still gotchas about figuring out if a .js is supposed to be one mode or the other.

On file extension: we add a character to your file, and make sure to update code editors etc. to know about the MIME.

On pragma / grammar change: we are similar to file extension (w/o MIME question), but won't get it through standards. Questions remain about ambiguity if not mandatory.

Vanuan:
> if consumers have to know or care about what module format I'm using, it's going to be a disaster for everyone

As a module consumer (I've never written any npm packages), I would rather be aware which module format I'm using than be bitten by unexpected bugs. For example, webpack supports AMD module format and it has already caused issues for me (leaking global).

If I'm maintaining an old codebase (`require`s all over the place), I would rather expect module authors to pubish a commonjs version if there's a will to support older version of node. If module author doesn't support new versions of node, I'd still be forced to not upgrade to a new node since there's no guarantee it will work the same.

If I'm mainaining a new codebase, I would rather expect new version of an npm package to contain es6 version of a module rather than having to append `-es` in every import line.

Would it be reasonable to expect a `--disable-commonjs` flag for people that want the cleanest and freshest version of node? Even if it means that their choice of libraries would be limited.

---

The above is from a consumer perspective. Producers would probably want their packages to work indefinitely without any maintenance.

bmeck:
web UI problem caused me to squash, old branch is at https://github.com/bmeck/node-eps/tree/rewrite-esm-bak

@targos please re-review

002-es-modules.md
```diff
@@ -0,0 +1,511 @@
...
+##### 3.3.1.1. HostResolveImportedModule Search
+
+Notes:
+
+* The CLI has a location URL of the process working directory.
+* Paths are resolved to realpaths normally *after* all these steps.
+
+1. Apply the URL parser to `specifier`. If the result is not failure, return the result.
+2. If `specifier` does start with the character U+002F SOLIDUS (`/`), the two-character sequence U+002E FULL STOP, U+002F SOLIDUS (`./`), or the three-character sequence U+002E FULL STOP, U+002E FULL STOP, U+002F SOLIDUS (`../`)
+  1. Let `specifier` be the result of applying the URL parser to `specifier` with importing location's URL as the base URL.
```
Jessidhia: github's markdown renderer is confused by 2-space indent... this renders as `3.` instead of an indented `1.` 😅 

maybe 4 spaces makes it happier?

002-es-modules.md
```diff
@@ -0,0 +1,511 @@
...
+3. Let `searchable` be the result of applying the URL parser to `./index` with `specifier` as the base URL.
+4. If it does not throw an error, return the result of applying the file search algorithm to `searchable`.
+5. Throw an error.
+
+##### 3.3.1.6. Directory Search
+
+1. Let `dir` be a new URL from `specifier`.
+2. If `dir` does not have a trailing `/` in its pathname append one.
+3. Let `searchable` be the result of applying the URL parser to `./package.json` with `dir` as the base URL.
+5. If the resource for `searchable` exists and it contains a "main" field.
```
Jessidhia: Missing step `4` (renumbered?)

002-es-modules.md
```diff
@@ -0,0 +1,511 @@
...
+  1. Let `main` be the result of applying the URL parser to the `main` field with `dir` as the base URL.
+  2. If it does not throw an error, return the result of applying the package main search algorithm to `main`.
+6. If it does not throw an error, return the result of applying the index main search algorithm to `dir`.
```
Jessidhia: The section this references is named just "Index search"

002-es-modules.md
```diff
@@ -0,0 +1,511 @@
...
+
+##### 3.3.1.7. Module Search
+
+1. Let `package` be a new URL from the directory containing the importing location.
+2. If `package` does not have a trailing `/` in its pathname append one.
+3. Let `searchable` be the result of applying the URL parser to `./node_modules/${specifier}` with `package` as the base URL.
+4. If it does not throw an error, return the result of applying the file search algorithm to `searchable`.
+5. If it does not throw an error, return the result of applying the directory search algorithm to `searchable`.
+6. Let `parent` be the result of applying the URL parser to `../` with `package` as the base URL.
+7. If it does not throw an error, return the result of applying the module search algorithm to `specifier` with an importing location of `parent`.
```
Jessidhia: `/../` is still a valid path (`..` in `/` just points to itself)

002-es-modules.md
```diff
@@ -0,0 +1,511 @@
...
+In the case that an `import` statement is unable to find a module, Node should
+make a **best effort** to see if `require` would have found the module and
+print out where it was found, if `NODE_PATH` was used, if `HOME` was used, etc.
+
+#### 4.4. Shipping both ESM and CJS
+
+When a `package.json` main is encountered, file extension searches are used to
+provide a means to ship both ESM and CJS variants of packages. If we have two
+entry points `index.mjs` and `index.js` setting `"main":"./index"` in
+`package.json` will make Node pick up either, depending on what is supported.
```
Jessidhia, line 275: I take it that there is no intent to support the ad-hoc `module` field.

(it's better that way, tbh; `module` is "specified" to require conversion to ES5 for example)

002-es-modules.md
```diff
@@ -0,0 +1,511 @@
...
+
+After *any* CJS finishes evaluation, it will be placed into the same cache as ESM.
+The value of what is placed in the cache will reflect a single default export pointing to the value of `module.exports` at the time evaluation ended.
+
+Essentially after any CJS completes evaluation:
+
+1. if there was an error, place the error in the ESM cache and return
+2. let `export` be the value of `module.exports`
+3. if there was an error, place the error in the ESM cache and return
+4. create an ESM with `{default:module.exports}` as its namespace
```
Jessidhia, line 338: Will this be a bound export, or will it snapshot the value of `module.exports` at caching time?

002-es-modules.md
```diff
@@ -0,0 +1,511 @@
...
+
+No existing code will be affected.
+
+#### 4.8.1. ES exports are read only
+
+The objects create by an ES module are [ModuleNamespace Objects][5].
+
+These have `[[Set]]` be a no-op and are read only views of the exports of an ES
+module. Attempting to reassign any named export will not work, but assigning to
+the properties of the exports follows normal rules.
```
Jessidhia: It's also not possible to add any extra keys

002-es-modules.md
bmeck: `/../` in URL parser becomes `/` in step 6.

002-es-modules.md
bmeck: no plan

002-es-modules.md
bmeck: snapshot right then and there, grabs the value just like step 4. does not make a getter or anything.

002-es-modules.md
```diff
@@ -0,0 +1,511 @@
...
+
+ESM imports will be loaded asynchronously. This matches browser behavior. This means:
+
+* If a new `import` is queued up, it will never evaluate synchronously.
+* Between module evaluation within the same graph there may be other work done. Order of module evaluation within a module graph will always be preserved.
+* Multiple module graphs could be loading at the same time concurrently.
+
+### 3.2. Determining if source is an ES Module
+
+A new file type will be recognised, `.mjs`, for ES modules. This file type will be registered with IANA as an official file type, see TC39 issue. There are no known issues with browsers since they [do not determine MIME type using file extensions](https://mimesniff.spec.whatwg.org/#interpreting-the-resource-metadata).
```
targos: Link for the TC39 issue?

002-es-modules.md
Jessidhia: It still won't throw an error, though, so this looks like it could loop forever (?)

002-es-modules.md
Jessidhia: So it's as if it was wrapped with a module that did `export default require('CJS')`

bmeck: roughly

002-es-modules.md
bmeck: Would fail in step 1 when it tries to grab the directory of `parent` which is `/`. Can clarify in step 1 by saying if the parent is same as importing location it should error.

targos:
What about direct execution of ESM code from the CLI? Should we specify this here?
I'm thinking about `node -e "import foo from 'bar'; foo();"`

bmeck:
@targos I've left the CLI flags out of this EP as it is not something I think belongs in the EP itself. The general consensus is that you must include a `--module` flag to change goal of source passed via argv or STDIN

bmeck:
landed in 6cc060e94e56859bdb446a0820ef4704731ff0a8

PR #41, opened 2016-09-03
Fix typo in es6 modules proposal

ethanresnick:
Hi,

I may be understanding the proposal wrong, in which case please close this PR.

If I'm understanding it correctly, though, I'd expect `import * as bar from './cjs.js';` to be a ModuleNamespaceObject rather than the `module.exports` function value, so name would be undefined. Right?


ethanresnick:
Whoops, just saw #39, so I guess this is irrelevant.

PR #42, opened 2016-09-04
proposal: `node inspect` CLI debugger

hybrist:
I think this came up during the original V8 inspector discussions.

P.S.: I created a [POC port of `lib/_debugger.js` to the `--inspect` protocol](https://github.com/buggerjs/node-inspect/blob/master/lib/node-inspect.js) to make sure this could reasonably be done in core without pulling in additional dependencies.

### DIFF
--- a/XXX-node-inspect-cli-debugger.md
+++ b/XXX-node-inspect-cli-debugger.md
@@ -1,0 +1,37 @@
+| Title  | `node inspect` CLI debugger |
+|--------|-----------------------------|
+| Author | @jkrems                     |
+| Status | DRAFT                       |
+| Date   | 2016-09-04                  |
+
+## Description
+
+For the old V8 debugger protocol,
+node offers two ways to use it:
+
+1. `node --debug <file>`: Start `file` with remote debugging enabled.
+2. `node debug <file>`: Start an interactive CLI debugger for `<file>`.
+
+But for the Chrome inspector protol,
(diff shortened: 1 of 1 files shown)


hybrist:
I published the "polyfill" to npm as `node-inspect` (requires node 6):

``` bash
npm install --global node-inspect
```

For the implementation status, see this [tracking issue](https://github.com/buggerjs/node-inspect/issues/1). Most things should work already.

jasnell:
This is definitely interesting but I'm wondering if it would be better to develop this as a standalone client for the time being. Once it's functional, then we can look to see how much sense it makes to bringing it in to core.

hybrist:
That definitely makes sense. And I'm not looking to submit the code I linked to above to core in the next few days or immediately after this proposal is finalized/merged.

But my impression was that the current command line debugger is already depending on (more or less) unsupported parts of V8. So it seemed like a good idea to have an official plan for "the day it breaks". The proposal explicitly doesn't mention my repo. It just suggests a potential interface for transitioning the official built-in debugger from the old V8 protocol to the new inspector protocol.

jasnell:
btw, relevant to this: https://github.com/nodejs/node/issues/7266

hybrist:
Thanks for the link! Must've been blind. :)

hybrist:
After playing around with the finished debugger - it feels like a wasted opportunity to go down this route.

PR #43, opened 2016-09-08
Node v2.0 modules

Fishrock123:
My, possibly transitional, "counter-proposal" to https://github.com/nodejs/node-eps/pull/39.

Mostly hashed out, some details may be missing as this style of thing is not my forte.
This is what is roughly described in a recent meeting with other CTC members.

It is driven by an interest of our existing ecosystem, to maintain as much compatibility with everyone's existing hard work as possible, while allowing newer functionality to be used without major consequence.

Like many other things in node core, it is a compromise. A compromise from idealism to better suit the reality of our platform.

cc @bmeck, @ljharb, @nodejs/ctc i guess

_Dear lord I hope this doesn't gather a ton of hate._

Edit: this is rebased on #39 for some language simplicity reasons. I can probably merge those directly into his commit though.


002-es-modules.md
```diff
@@ -0,0 +1,175 @@
...
+  - The Reflective Module Record is described below as the `ModuleRecord` type.
+2. Allow Node.js modules to use "Module Mode" without using Async Module Resolution.
+
+### 1.1. Avoidance of the Reflective Module Record
+
+Note: The RMR is only a collection of direct pointers in a JIT'd environment.
+
+The RMR poses several interoperability and tooling problems:
+
+1. Importing from NCJS has no support for Named Imports.
```
ljharb, line 39: this is unfortunate :-( i think it's an OK thing to sacrifice, but basically everyone using babel right now to named-import from CJS is going to break.

Fishrock123: Note: this is a problem with using the RMR which this proposal avoids, these are not problems with this proposal. :) 

(Also saying for anyone else reading this.)

002-es-modules.md
```diff
@@ -0,0 +1,175 @@
...
+
+### 3.2.1 Function Wrap Approach
+
+To achieve this, modules would be wrapped in a function like a regular script
+file, but evaluated as a Module under the Module Parse Goal. However, that
+wrapper _may_ be exported with an `export` statement under the hood. As such,
+it _may_ look like the following:
+
+```js
+export default function (exports, require, module, __filename, __dirname) {
```
Jessidhia, line 159: So, internally, the execution engine would do (the runtime equivalent of) `import moduleBody from 'required-module'; moduleBody(exports, require, module, __filename, __dirname); return exports;`?

002-es-modules.md
```diff
@@ -0,0 +1,175 @@
...
+
+* implement interoperability for the Module Parse Goal and Node's existing
+module system
+
+## 1. Purpose
+
+1. Allow Node.js modules to use "Module Mode" without hitting the Reflective
+Module Record.
+  - The Reflective Module Record is described below as the `ModuleRecord` type.
+2. Allow Node.js modules to use "Module Mode" without using Async Module Resolution.
```
hybrist: Could you clarify what using "Module Mode" means? From what I can tell the only difference between using NM2 over NCJS would be that the former runs in strict mode..? But it can neither import nor export and top level code doesn't run in a module scope. So... wouldn't adding a `"use strict"` to the existing module wrapper have the same (observable) effect?

002-es-modules.md
Fishrock123: I .... don't think so? I don't follow.

002-es-modules.md
Fishrock123: Module Mode means to execute using the Module Parse Goal which has implicit strict mode, no `with`, new keywords, etc.

> So... wouldn't adding a `"use strict"` to the existing module wrapper have the same (observable) effect?

Absolutely not.

002-es-modules.md
ljharb: Ah, understood. Thanks for clarifying!

002-es-modules.md
```diff
@@ -0,0 +1,175 @@
...
+The `.mjs` file extension will be searched for prior to any `.js` file in any
+algorithm that uses searching (`require`); e.g. once the Node
+resolution algorithm reaches file expansion, `path + '.mjs'` would be searched
+prior to `path + '.js'` when performing `require(path)`.
+
+# 3.1.1. Inter package loading using file extension breakage.
+
+There is knowledge of breakage for code that upgrades inner package
+dependencies such as `require('foo/bar.js')`. As `bar.js` may move to
+`bar.mjs`. Since `bar.js` is not the listed entry point this was considered
```
ljharb, line 142: Similarly, a common community "best practice" (although not a universal one) is to always omit extensions when requiring - to allow for "bar.js" to become "bar/index.js" transparently, for example - and `require('foo/bar')` would not break under this proposal.

002-es-modules.md
```diff
@@ -0,0 +1,175 @@
...
+acceptable.
+
+### 3.2. Importing Modules in NM2
+
+Node v2.0 Modules do not implement `import` and `export`.
```
ljharb: If this is the case, then isn't it not the Module parsing goal?

002-es-modules.md
```diff
@@ -0,0 +1,175 @@
...
+Like normal scoping rules, if a variable does not exist in a scope, the outer scope is used to find the variable. Since NM2 are always strict, errors may be thrown upon trying to use variables that do not exist globally when using NM2.
+
+### 4. Caveats
+
+This approach saves existing operability and tooling functionality in exchange
+for a couple caveats, namely:
+
+1. `import` & `export` are unimplemented / unavailable to users
+2. Top-level scope is unavailable.
+  - As such, top-level `await` is unavailable. (Also for sync safety reasons.)
```
ljharb: this seems like it precludes valuable use cases in app entry points :-/

002-es-modules.md
hybrist: > Absolutely not.

I assumed as much! I'm really just looking for maybe a few concrete examples of the differences. Without those it's pretty hard to tell what the practical differences between NM2 and NCJS modules are. Especially for people who only watch the discussion from the distance (like myself) and aren't deeply involved. :)

ljharb: @jkrems in the Script goal, `import` and `export` are not permitted, `await` is not a keyword, strict mode is not automatic - to name a few.

002-es-modules.md
Jessidhia: What is the `export default` there for?

Fishrock123: > However, that wrapper _may_ be exported with an `export` statement under the hood. As such, it _may_ look like the following:

002-es-modules.md
Fishrock123: This was copied over from @bmeck's proposal and probably beong in that thread?

002-es-modules.md
Fishrock123: It still runs in the VM's Module Parse Goal mode, so it technically fits within the Module Parse Goal even if certain features are... inaccessible.

002-es-modules.md
Fishrock123: Could you expand on that?

If we want to synchronously resolve, it is not safe. (Although we could / do block, maybe it's not an issue.)

Also, if we function wrap.... it's just not available.

We could wrap in an async function but that also gets into the whole unsafe sync promise resolution thing.

Fishrock123: Honestly though, it doesn't seem that valuable to me since this is already going to be possible by just making an async function and sticking stuff in it.

Having an extra layer of indentation pales in comparison to RMR problems imho.

002-es-modules.md
hybrist: Yes, but further down this proposal excludes `import` and `export` from NM2 (because all code is in a closure, it can't access `import`/`export`). So other than strict mode that leaves "`await` is a keyword"? Trying to understand why I would use NM2 over NCJS if that's the only two differences.

002-es-modules.md
ljharb: I don't agree that's a claim that can be made - if `import` is not accessible at the top level, it's not a module. I understand the parallels to CJS and top-level not being available, but I don't think that's a good idea nor a mistake that should be re-made.

ljharb: Another point - at some point, sooner than later, new JS API features may _only_ be made available via `import` - ie, importing "builtin" modules. If `import` is excluded, then `node` would have to do something like overload `require` to support builtin modules, which has its own problems.

002-es-modules.md
ljharb: It's not about layers of indentation, it'd be about an exception terminating the process, which top-level await allows, but an async function would not.

002-es-modules.md
hybrist: Just skipped through the spec and tried to find other things that depend on the goal symbol (other than `await`) and came up empty. :( Is there a list somewhere?

Fishrock123: Future ES features will only target the Module Parse Goal.

hybrist: > Future ES features will only target the Module Parse Goal.

Okay, so trying to pin this down: There wouldn't be any concrete reason _yet_ for somebody to use NM2 over NCJS. But there are signals from TC39 that at some point in the future there might be proposals for JavaScript that would require using the Module Parse Goal? That feels pretty vague.

002-es-modules.md
Fishrock123: > it'd be about an exception terminating the process

The exception would bubble as usual?

ljharb: Not inside an async function - it's identical to an unhandled rejected promise, and the same lack of a solution there applies to non-top-level await.

dinoboff:
I don't really see the point.

If async loading is such an issue, I would prefer to have CJS scripts not able to load modules than to have half implemented ES modules. ES module developers would still need to publish a transpiled version but they would be able to run their module natively.

002-es-modules.md
bmeck: @jkrems agree on vague but thats a lot of what we are working with.

002-es-modules.md
bmeck: will re-add suggested workaround to rewrite

Fishrock123:
@dinoboff See the discussion at https://github.com/nodejs/node-eps/pull/43#discussion_r77944802

In summary, new ES Features will likely only target the Module Parse Goal, and as such it is a good idea to make it available, even if it is not within ES Modules due to problems with them.

> half implemented ES modules

Hmmm 💭 I don't think this ever claimed to be that.

This claims to implement the Module Parse Goal while specially not implementing ES Modules in the loading / import / export sense, since there is many problems with that.

002-es-modules.md
Fishrock123: > but I don't think that's a good idea nor a mistake that should be re-made.

You think top-level in the script goal is a mistake?

I think top-level `return` is useful enough in Node to make that "mistake".

> Another point - at some point, sooner than later, new JS API features may _only_ be made available via `import` - ie, importing "builtin" modules. If `import` is excluded, then `node` would have to do something like overload `require` to support builtin modules, which has its own problems.

Ah right.

I have a second idea, and will update the EP shortly with a second (less favorable imo) implementation option.

Edit: second idea may not be viable..

002-es-modules.md
Fishrock123: @ljharb does top-level `await` block?

Or are you talking about immediate exceptions?

bmeck: @Fishrock123 it suspends execution, the exact mechanics are _presumed_ to be the same suspension until Promise resolution as an async function.

002-es-modules.md
bmeck: node could also special case those in `require()` and defer to an intermediate internal module to grab the values pretty easily.

002-es-modules.md
ljharb: I'm saying that exceptions do not bubble inside async functions just like they don't in Promises. Top-level await, however, would have that functionality, because it's not inside an async function.

Fishrock123:
Updated with some more problems with Spec ES Modules and addressed some comments.

002-es-modules.md
Fishrock123: After some discussion with @bmeck I don't think this could be safely done from a not-fully-async import.

Fishrock123:
Also a note which I don't know where to put yet: This retains "top-level" `return`, which is important in many Node scripts.

ljharb:
That's not part of the module goal though. There isn't supposed to be a top-level return, and I'm not sure why that's valuable in CJS modules anyways.

Fishrock123:
Updated with an example for `### 1.2. Avoidance of Async Module Resolution`.

dinoboff:
@Fishrock123 Node could support three types:
- cj scripts - parsed in strict mode using legacy loader.
- cj modules - parsed in module mode using legacy loader.
- es modules - parsed in module mode using System loader.

With the two loaders supporting or not each other.

This proposal doesn't need to be a rewrite of 002-es-modules; why not a 004-cj-modules proposal?

Fishrock123:
@dinoboff unsure if that would be useful.

Qard:
If we can get top-level await, we could just do something like `await Promise.resolve(exports)` _inside_ the require function, making the asynchrony transparent to existing modules with the sync expectation.

bmeck:
@qard not sure I understand, `await` would still only be available within _async_ contexts (Async functions / Module scope). Any context using it would need to provide an async wrapper (Promise) as `await` only pauses the current stack frame.

Qard:
I understood top-level async/await as having await _everywhere_ and, when used outside an async function, would halt the entire JS thread. Perhaps I misunderstood it, but it seemed like a lot of people had the same understanding of it.

Alternatively, a function on promises to block until resolution would help too, like a thread join. Though I'm not sure how that'd impact _other_ things in the microtask queue, and I somehow doubt TC39 would be into that idea...

loganfsmyth:
> I understood top-level async/await as having await everywhere and, when used outside an async function, would halt the entire JS thread. Perhaps I misunderstood it, but it seemed like a lot of people had the same understanding of it.

`top-level await` is specifically about allowing `await` in the "top level" scope of a module (outside all functions), it would still never be usable inside arbitrary functions. If a module is mid-evaluation and hits an `await`, it will suspend execution of the module body until the promise resolved, but the JS thread would not be halted and other JS could be executed while the module body was waiting, the same way an async function would.

Jessidhia:
There could be the option of having a syntax extension where a call to `require` would have an implicit `await` in front of it.

It would still not be completely transparent, though, as it'd break `require`s not called at the top level or in `async` functions.

Qard:
Ah, in that case...IGNORE ME.

Fishrock123:
closing, agreement has come around to other options

PR #48, opened 2017-01-17
proposal: JS tracing APIs

jasongin:
@nodejs/diagnostics and anyone else interested, please review this proposal draft.

### DIFF
--- a/00X-tracing.md
+++ b/00X-tracing.md
@@ -189,14 +189,14 @@ interface Tracing {
 }
 ```
 
-### 2.4 Emit API overloads
+### 2.4 Event-type-specific emit APIs
 The `emit()` method defined above mostly follows the familiar `EventEmitter` pattern for emitting
-events as objects. However some additional overloads are proposed that take separate parameters
-instead of an event object. Using these overloads, the four emitted events in the previous code
-example can be rewritten more concisely:
+events as objects. However some alternative methods are proposed for each type of event, that take
+separate parameters instead of an event object. Using these methods, the four emitted events in the
+previous code example can be rewritten more concisely:
 
 ```javascript
-tracing.emit(category, 'myEvent');
+tracing.instant(category, 'myEvent');
 
 const scopeId = tracing.begin(category, 'myTask');
 tracing.end(category, 'myTask', scopeId);
@@ -204,15 +204,13 @@ tracing.end(category, 'myTask', scopeId);
 tracing.count(category, 'myCounter', 3);
 ```
 
-These overloads are slightly more efficient because they doesn't require creation of an object for
+These methods are slightly more efficient because they doesn't require creation of an object for
 the event. And for simple cases they are easier to write and read.
 
 ```typescript
 interface Tracing {
   /**
-   * Emits a tracing event of type "instant" using an event name and optional event args. For
-   * simple "instant" events, this is slightly more efficient than the other emit() overloads
-   * because it doesn't require creation of an object for the event.
+   * Emits a tracing event of type "instant" using an event name and optional event args.
    *
    * @param category Required tracing category name or array of one or more category names for
    *   the tracing event.
@@ -223,7 +221,7 @@ interface Tracing {
    * @returns True if the event was emitted; false if the event was not emitted because tracing
    *   was not enabled for any of the event categories.
    */
-  emit(
+  instant(
     category: string | string[],
     name: string,
     args?: { [name: string]: any } | () => { [name: string]: any }): boolean;
@@ -331,7 +329,7 @@ tracing.on(['example1', 'example2'] , e => { // e is a TracingEvent object
 tracing.emit(['example2', 'example3'], 'info', 'Hello tracing!');
 ```
 
-### 3.2 Listern API specification
+### 3.2 Listener API specification
 ```typescript
 interface Tracing extends CategoryEventEmitter ...
 
--- a/00X-tracing.md
+++ b/00X-tracing.md
@@ -1,0 +1,581 @@
+| Title  | JS Tracing APIs |
+|--------|-----------------|
+| Author | @jasongin       |
+| Status | DRAFT           |
+| Date   | 2017-01-12      |
+
+## 1. Overview
+This document proposes adding new JavaScript APIs for high-performance tracing, built on top of
+the native trace event framework from Chromium that is already integrated into the Node engine.
+
+There are three major parts to this proposal:
+  * Tracing emitter APIs
+  * Tracing listener APIs
+  * Console tracing API extensions
+
+The listener APIs and console API extensions are optional: either or both of those could
+conceivably be rejected while still accepting the other part(s).
+
+### 1.1 Justification
+There are benefits to having tracing APIs built-in to the Node platform, compared to implementation
+as an external module:
+ * **Holistic view** - By leveraging the same tracing system across all layers of the stack
+   (v8, Node, libraries, and applications), diagnostic tools can offer a complete and
+   integrated view of the system.
+ * **Standardization** - If high-performance tracing APIs are built-in to Node core, then popular
+   third-party libraries are much more likely to instrument their code using the the APIs. As the
+   ecosystem standardizes on a common system for tracing, applications developers benefit from
+   improved diagnostic capabilities.
+ * **Performance** - High performance is generally achieved by doing as much of the tracing work
+   as possible in native code and off of the main thread. While much of that could be accomplished
+   using an external native module, there may be some additional optimizations that can be done
+   only through integration with the GC and/or JIT.
+
+### 1.2 Background
+The [Node.js Diagnostics WG](https://github.com/nodejs/diagnostics) became aware of a
+high-performance tracing framework being integrated from Chromium into v8. They agreed that the
+Node.js engine could also benefit from using the same tracing framework. See the
+[meeting notes from initial tracing discussion](https://github.com/nodejs/diagnostics/blob/master/wg-meetings/2016-06-01.md)
+and the [related WG issue](https://github.com/nodejs/diagnostics/issues/53).
+
+Integration of the C++ trace event framework into Node was
+[completed recently](https://github.com/nodejs/node/pull/9304). It can be
+[enabled at runtime via node command-line options](https://github.com/nodejs/node/blob/master/doc/api/tracing.md).
+
+Note the Node engine (C++ code) is not yet *instrumented* with calls to the tracing macros. That
+work is planned, but is outside the scope of this proposal.
+
+## 2. Tracing emitter APIs
+The tracing emitter APIs enable JavaScript code to emit tracing events of any of the supported
+types along with associated event metadata. The APIs are primarily intended to be used in the
+following scenarios:
+ * Instrumentation of application code, enabling application developers to collect performance
+   and diagnostic tracing data about their application code.
+ * Instrumentation of library code, enabling developers of applications consuming the library
+   to collect performance and diagnostic tracing data about the library, either separate from or
+   combined with application and engine tracing.
+
+The tracing emitter APIs might also be used as a low-level output format for higher-level tracing
+or logging frameworks. For example, logging frameworks that enable pluggable "sinks" for log events
+may support Node.js tracing as another sink option. However, since these APIs are primarily
+designed for performance tracing, they are not optimized for logging scenarios. For example,
+they don't define any log-level metadata or filtering based on it.
+
+### 2.1 New core module
+The proposed new tracing emitter APIs (and listener APIs) belong in a new Node.js core module:
+```javascript
+const tracing = require('tracing');
+```
+The "tracing" module name is [already reserved](https://www.npmjs.com/package/tracing) for this
+purpose. Alternatively, "trace" might be a good name, however that would have to be reconciled
+against the existing ["trace" npm package that generates long stack traces using AsyncWrap](https://www.npmjs.com/package/trace).
+
+### 2.2 Example use
+```javascript
+// Enable recording for the 'example' category. (Can also enable via node command-line.)
+const category = 'example';
+tracing.enableRecording(category);
+
+tracing.emit(category, { eventType: 'instant', name: 'myEvent' });
+
+const scopeId = tracing.emit(category, { eventType: 'begin', name: 'myTask' });
+tracing.emit(category, { eventType: 'end', name: 'myTask', id: scopeId });
+
+tracing.emit(category, { eventType: 'count', name: 'myCounter', value: 3 });
+```
+
+### 2.3 Emitter API specification
+APIs are specified in TypeScript syntax for clarity in this doc. Actual implementation is of course
+plain JS.
+
+```typescript
+/**
+ * Defines the valid string values for the eventType property of a TracingEvent object.
+ */
+type TracingEventType = "instant" | "begin" | "end" | "count";
+
+interface TracingEvent {
+  /**
+   * Required type of event; must be one of the allowed event type string values.
+   */
+  eventType: TracingEventType;
+
+  /**
+   * Event category or category list. May be omitted if the TracingEvent object is passed to
+   * Tracing.emit() where the category is specified as a separate parameter.
+   */
+  category?: string | string[];
+
+  /**
+   * Name of the event (not necessarily unique). Required for all event types.
+   */
+  name: string;
+
+  /**
+   * Integer identifier that can be used to correlate paired "begin" and "end" events or multiple
+   * "count" events and distinguish them from other timings or counters that use the same name.
+   * For a "begin" event a scope identifier is auto-generated and must be supplied again with
+   * the corresponding "end" event. For "count" events a counter identifier is optional. This
+   * property is not used with "instant" events.
+   */
+  id?: number;
+
+  /**
+   * Required counter value for "count" events; not used for other event types. This must be either
+   * a number, a dictionary of name-number pairs, or a function that returns one of those. If a
+   * function, it will be invoked to retrieve the value(s) only if tracing is enabled for the
+   * category. (The tracing system currently requires multi-value counters to have exactly two
+   * values.)
+   */
+  value?: number | { [name: string]: number } | () => (number | { [name: string]: number });
+
+  /**
+   * Optional arguments for "begin", "end", and "instant" events. If specified, this must be a
+   * either a dictionary of name-value pairs or a function that returns a dictionary. If a
+   * function, it will be invoked to retrieve the args only if tracing is enabled for the category.
+   * Not used for "count" events.
+   */
+  args?: { [name: string]: any } | () => { [name: string]: any };
+}
+
+interface Tracing {
+  /**
+   * Checks whether tracing is enabled for a category or for any categories in an array. Tracing
+   * is enabled for a category if either recording is enabled for the category or there are one
+   * or more listeners for the category.
+   *
+   * @param category Required tracing category name or array of category names.
+   * @returns True if tracing is enabled for the category or for any of the categories in the
+   *   array.
+   */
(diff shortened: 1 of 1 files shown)


00X-tracing.md
```diff
@@ -0,0 +1,581 @@
...
+// Enable recording for the 'example' category. (Can also enable via node command-line.)
+const category = 'example';
+tracing.enableRecording(category);
+
+tracing.emit(category, { eventType: 'instant', name: 'myEvent' });
+
+const scopeId = tracing.emit(category, { eventType: 'begin', name: 'myTask' });
+tracing.emit(category, { eventType: 'end', name: 'myTask', id: scopeId });
+
+tracing.emit(category, { eventType: 'count', name: 'myCounter', value: 3 });
```
Qard, line 84: This emit function is different enough from EventEmitter.emit that I'd prefer it was named differently so as to not confuse new users.

00X-tracing.md
```diff
@@ -0,0 +1,581 @@
...
+   * @param args Optional dictionary of name-value pairs, or a function that returns a dictionary.
+   *   If a function, it will be invoked to retrieve the args only if tracing is enabled for the
+   *   category.
+   * @returns True if the event was emitted; false if the event was not emitted because tracing
+   *   was not enabled for any of the event categories.
+   */
+  emit(
+    category: string | string[],
+    name: string,
+    args?: { [name: string]: any } | () => { [name: string]: any }): boolean;
```
Qard: Maybe we should call this instant rather than emit. Multi-signature methods risk unforeseen problems, as the buffer issue has proven.

00X-tracing.md
Qard: Perhaps we should support *only* the alternate API? All the conditional validation needed for this unified function is a bit worrying.

00X-tracing.md
```diff
@@ -0,0 +1,581 @@
...
+listeners.
+
+Relatively few applications and libraries are expected to use the listener APIs (as compared
+to the much broader applicability of the emitter APIs). However, the ability to listen to tracing
+events could be very valuable to certain application-monitoring, profiling, and diagnostic tools
+and SDKs.
+
+To support listeners, the `Tracing` class (from the previous section) extends a new
+`CategoryEventEmitter` base class. That class is similar to `EventEmitter`, but does not extend
+`EventEmitter` because it has different behavior related to the multi-category support.
```
Qard, line 302: Again, similar yet different enough from a typical event emitter that I worry it'd be confusing to new users.

00X-tracing.md
```diff
@@ -0,0 +1,581 @@
...
+// The listener receives events emitted for *any* matching category.
+tracing.on(['example1', 'example2'] , e => { // e is a TracingEvent object
+  if (e.name === 'info') {
+    console.log('Got a tracing event: ' + e.args.message);
+  }
+});
+tracing.emit(['example2', 'example3'], 'info', 'Hello tracing!');
+```
+
+### 3.2 Listern API specification
```
Qard: Listen*

AndreasMadsen:
Is this a follow up on "trace events" as introduced in https://github.com/nodejs/node/pull/9304?

---

I'm generally against using the term _tracing_ formally, it can mean so many things. For example we already have:
* LTTng (Linux), SystemTap (Linux), DTrace (OSX), ETW (Windows)
* `async_hooks`
* `trace_events`

all of which does some kind of _tracing_. If this API integrates with `trace_events` I think we should just call the module `trace_events`.

00X-tracing.md
```diff
@@ -0,0 +1,581 @@
...
+## 4. Console tracing API extensions
+The JavaScript `console` module (in Node.js and the web) already defines a set of basic tracing
+methods that align well with the four tracing event types described above. These methods are NOT
+part of any official standard, though there is
+[a working group that maintains a spec](https://github.com/whatwg/console)
+for the purpose of encouraging interoperability.
+
+The existing Node.js implementation of those Console tracing methods simply logs tracing events to
+the console. This proposal extends those methods with a new optional category parameter to enable
+emitting categorized tracing events in a way that is backward-compatible with existing console APIs.
```
Qard, line 426: I don't like the idea of deviating from the behaviour of the console API spec. If they decide to add their own extra argument in the future, suddenly we have problems.

jasongin:
Yes, the Background section of the proposal doc links to that "trace events" PR.

As also noted in the doc, the module name "tracing" was already reserved for this purpose long ago. I don't feel strongly about it personally, certainly we can go with whatever name is agreed by consensus.

00X-tracing.md
othiym23: In addition, the polymorphism of multi-signature methods has implications for optimization, which raises performance concerns. Historically, this has led to problems for APM and other diagnostic solutions that are intended to be embedded in production applications. In general, the simpler / closer to monomorphic the APIs can be made, the better.

othiym23:
While I agree that treating those three APIs as separate concerns that can be implemented separately, I think it's helpful to treat emitting and listening of trace data as duals for a single abstraction.

As someone who, like @Qard, has worked on APM products in the past, the challenge for me was that there have been a variety of strategies for emitting trace events, but no reliable way to observe that data across platforms and deployment environments (in particular, PaaS and containerization). In practical terms, it feels like it's a lot easier to hit performance and overhead targets for the emitting rather than the listening side, and coming up with an efficient in-process interface for capturing that data would be the biggest win for a whole class of use cases.

00X-tracing.md
jasongin: Agreed, this method would be better named `instant`. (Or perhaps `timeStamp`?)

00X-tracing.md
jasongin: I considered trying to make it more different from `EventEmitter`, but it felt silly to use different names for operations that are really the same as the normal `EventEmitter` methods.

I think the differences probably should not really be a problem. You can in fact use a `CategoryEventEmitter` (which the `Tracing` module extends) just like a normal `EventEmitter` if you don't care about the multi-category support.

Would it help to think of `CategoryEventEmitter` as a *subclass* of `EventEmitter`? (I considered literally extending the class, but it didn't provide any benefit since all the methods needed to be overridden to use different data structures from the base class.)

00X-tracing.md
jasongin: Understood; I haven't entirely convinced myself that it is a good idea either. That's why I made it an optional part of the proposal. It might be nice for some scenarios, so I'd like to at least get some feedback about the idea of integrating with the console tracing APIs.

00X-tracing.md
hybrist: +1 for `instant` because it matches the terms used in the trace format descriptions. It makes it easier to find it elsewhere.

jasongin: Renamed to `instant` in a new commit.

00X-tracing.md
jasongin: Fixed in a new commit.

trevnorris:
I'd like to see documentation about functionality this introduces that would be impossible to implement on top of async hooks (the async hooks EP is up to date).

jasongin:
I will update the doc to explain how this proposal relates to async hooks. I expect tracing instrumentation of node core _can_ leverage async hooks for tracing of async operations, so in that regard the features are complementary.

Async hooks provide a way to hook into the beginning and end of async operations built on libuv primitives. Of course those async operations are a key subset of performance-related events.

This proposal is about a way to capture and/or listen to all kinds of performance-related events, including non-async begin/end, point-in-time, and counters. Additionally and most importantly it provides a very high-performance way to write the events out to a trace log file.

targos:
Is this somehow related to the browser's User Timing API? See https://developer.mozilla.org/en-US/docs/Web/API/User_Timing_API

jasongin:
@targos It is not directly related, but I'm aware of it. I think it would not be a good fit to implement the User Timing API directly in node, because it only has a subset of the functionality (does not support counters, categories, or any mechanism for turning off the tracing) and it does not have the right event-listener model.

If desired, it should be straightforward to create a polyfill for the User Timing API that would use this proposed Node.js tracing API underneath.

trevnorris:
@jasongin I should have clarified. I specifically meant in regards to the JS API. I'm fine placing the trace event hooks in native source. I already have for some testing and they work great. My concern is the JS API. Even if the call does nothing but return immediately there's still enough overhead to be noticeable. Especially when the function call is cross module (V8 has a harder time optimizing those for some reason).

For async hooks to incur the least amount of overhead possible in similar situations I had to do some things I'm rather not proud of (see https://git.io/vDJq9 as an example). So I'd like to avoid placing anything directly in JS. I'd like to see if the JS trace events could be optional async hooks that are enabled when those specific trace events are requested.

jasongin:
The main purpose of a JS API for tracing is to enable people to _instrument their JavaScript code_. While instrumenting internal async handles and other internal constructs may also provide useful data, that is only a piece of the performance picture and is not the focus of this proposal.

> Even if the call does nothing but return immediately there's still enough overhead to be noticeable.

Only if your definition of "noticeable" is _extremely_ generous. When tracing is completely off, any of the tracing event-emitter functions will literally be a no-op function, the equivalent of:
```
tracing.emit = function () {}
```
Even if the JIT cannot completely optimize-out the function call, it is still going to be extremely low-cost, I think small enough to not be noticeable in any reasonable sense.

Then of course when tracing gets enabled, either by turning on recording or adding a listener for any category, the emit functions will get replaced with functions that need to do a little bit of work, but are still fast enough not to have a significant impact on most applications.

trevnorris:
I spent around 1.5 week figuring out the best way to optimize for calls to timers and nextTick in my async hooks PR. The reason it was so important to improve is because node doesn't code for the average performance case. We code it for the fast case. Which means measuring against a high volume scenario.

Take a simple http server and add hooks when resources are created and also add timing around each resource's callback. In a high volume scenario, let's say a proxy (and these numbers are small compared to what F5 pushes through node), and do 45k req/sec. Just tracking the mentioned locations you're going to add ~1.5 million calls/sec. That's ~1.5 million noops in JS that have now been added. Which aren't actually noops because V8 can't completely optimize them out. At that volume I'm not being "_extremely_ generous". I'm being practical.

Looking at the code from the link I posted you'd see that all the `emitBefore()`/`emitAfter()` calls are conditionally called. Because simply calling the noop function cost too much when I was benchmarking. Short of this is, I didn't spend a month tweaking every corner of the async hooks API to make sure it had no noticeable affect on core's performance just so we could throw in another similar API that would.

jasongin:
I appreciate the passion for performance. I am working on some benchmarks involving tracing in realistic scenarios. I will follow up when I have data ready to share.

00X-tracing.md
```diff
@@ -0,0 +1,579 @@
...
+purpose. Alternatively, "trace" might be a good name, however that would have to be reconciled
+against the existing ["trace" npm package that generates long stack traces using AsyncWrap](https://www.npmjs.com/package/trace).
+
+### 2.2 Example use
+```javascript
+// Enable recording for the 'example' category. (Can also enable via node command-line.)
+const category = 'example';
+tracing.enableRecording(category);
+
+tracing.emit(category, { eventType: 'instant', name: 'myEvent' });
```
naugtur, line 79: could `tracing` expose constants to pass in as eventType instead of expecting particular strings?

00X-tracing.md
mcollina: I agree, this should not be called `emit` at all. Maybe `record`?

00X-tracing.md
```diff
@@ -0,0 +1,579 @@
...
+  /**
+   * Emits a tracing event using an event object, with categories specified separately.
+   *
+   * @param category Required tracing category name or array of one or more category names for
+   *   the tracing event. Overrides any categories specified in the event object.
+   * @param e The event to be emitted.
+   * @returns Nonzero event id (either provided or generated) if the event was emitted, or null if
+   *   the event was not emitted because tracing was not enabled for any of the event categories.
+   */
+  emit(category: string | string[], e: TracingEvent): number | null;
```
mcollina, line 188: I think this is redundant, maybe we can try to avoid overloading as much as possible as it is harder to optimize, and this API should try to be as fast as possible.

00X-tracing.md
```diff
@@ -0,0 +1,579 @@
...
+```typescript
+interface Tracing extends CategoryEventEmitter ...
+
+/**
+ * Enables emitting and listening to events in multiple categories.
+ *
+ * This interface has methods that are similar to EventEmitter, with an important difference:
+ * events may be emitted under multiple categories simultaneously instead of a single event name,
+ * and listeners may listen to multiple categories while still only being called once per event.
+ */
```
mcollina, line 342: If it is similar but different, it should not be called `EventEmitter`, but something else.

00X-tracing.md
mcollina: I'm 👎  in adding any complexity to `console`.

00X-tracing.md
```diff
@@ -0,0 +1,579 @@
...
+
+### 6.1 Output serialization
+Currently, logs are simply JSON-serialized to a file; this JSON format is
+[understood by the Chrome dev tools](https://www.chromium.org/developers/how-tos/trace-event-profiling-tool). The C++ layer of the
+tracing framework supports pluggable exporters, for example Chromium has an
+[ETW exporter for Windows](https://chromium.googlesource.com/chromium/src/+/master/base/trace_event/trace_event_etw_export_win.h),
+which way may want to adapt to Node.js eventually. And there has been discussion of a more compact
+binary format. This is an area that needs some further development, however for performance reasons
+the tracing output probably would not be exposed to JavaScript anyway, beyond the listener APIs
+that are proposed here.
```
mcollina, line 562: Some time should be taken to design a processing pipeline for these. I think some tools and documentation to explain how to process this data format. Without interactive and easy-to-understand visualisations, tracing data is extremely hard to understand.

I should be straightforward to plot this data into a flamegraph for example.

mcollina (changes requested):
I'm waiting for this, but I think there are still some nits to be fixed in this EP.

rvagg:
﻿https://github.com/brycebaril/transaction-tracer is a proposal that @brycebaril put forward a few years ago coming out the Tracing Summit held a few years ago. Might be good to connect to some of that work. @othiym23 and @trevnorris you were involved in that too weren't you? Was there any other documentation or notes that came out of that which might be useful?

othiym23:
I'm pretty sure @brycebaril's notes were the most complete; @groundwater acted as our facilitator, and may have recorded some of the outcomes, but I don't know how extensive his notes were, given that he was facilitating. @tjfontaine and @misterdjules were our hosts at the Joyent Vancouver offices and had extensive contributions to make around our efforts to come up with something that merged both APM and dynamic tracing along the lines of DTrace and ETW. I believe @Qard was also at that meeting as well, representing his APM-vendor employers of the time (sorry, Stephen, I'm blanking on what company that was temporarily).

I don't think @trevnorris made it to that particular tracing meeting, although he was certainly integral to the discussions that @groundwater, @tjfontaine, and I were having earlier that year. Trevor _was_ at the error-handling "summit" convened by @dshaw earlier that year, at which @davepacheco was also present. As Trevor's involvement in those discussions eventually produced `async_hooks` and `AsyncWrap`, those APIs are far more concrete than any meeting notes are likely to be. ;)

I'll double-check my notebooks from the time, but my recollection is that I didn't keep detailed notes, in part because the discussion was largely focused on establishing the problem space more than solutions, in part because I'd moved on from APM at the time and was more just trying to act as a kind of expert witness.

misterdjules:
FWIW, my notes from the meeting were:

## General observations

- People don't want to have to monkey patch node to implement tracing.

- People seem to be more keen to keep the tracing facilities in node itself
rather than use platform tracing facilities like dtrace/ETW/{System,K}tap.

## What some APM providers do/need

### New Relic:
  * Problems they try to solve:
     - help developers make their applications faster.
     - Identify individual transactions that are slow.
     - Don't generate very broad info (no big call graph), but more targetted
     info.
     - As lightweight as possible. Needs to not have performance impact on the
     running app.

  * Solution they use:
    - Uses AsyncListener polyfills or core implementation + continuous local
    storage to tie things together.

### Strongloop:
  * Performance tracing: find performance problems and/or bugs (exceptions,
  leaks of descriptors, etc.).
    - Needs to know start and stop of async events.
  * Solution they use: monkey patch everything.
  * Other problems they try to solve: heap/cpu profiling.
  * Use or would like to use zones.
  * Mentioned being interested in other discrete events like GC or Node about to
  exit/not being able to exit, etc.

### Appneta:
  * Goal: time profiling of the whole async flow.
  * Mentioned needs to trace async calls in native modules.

### Concurix:
  * Also instrument synchronous code.
  * Light monkey patching.
  * Needs to not have performance impact on the running app.


## General comments:
  * Tools to allow native modules to emit tracing/performance events when they
  perform async operations using plain libuv.

## Conclusion of the "problem discussion"
  * Priorities:
    1. Unavoidable requirements:
      * Production ready.
      * Do not monkey patch node.
      * Do not alter behavior.
    2. Async tracing.
    3. Track async lifecycle.
    4. native module API.

## Implementation

### Concerns/Questions:
  - Keep a global state of async trees/transactions but how do we know when we
  can cleanup some of these long lasting state structures?

### Observations:
  - Have a simpler solution that just bubbles up simple information and not
  complete async trees. AL is too high level.

### Proposed API by TJ/Jacob (for a simple solution):
  * Terminology:
    - Namespace, provider, proble. Example:
    node:http:{client-request,client-response,etc.}
  * Advantages: similar to already defined probes.
  * API: global someProbe = createProbe('probeName', description of info that is
  going to be passed). The description of data is needed to support platforms
  providers like dtrace.
        `someProbe.enable()`
        `someProbe.fire(function () { return [ data that is passed to the callback of the consumer]})`
        `.list()`: returns all the probes/namespaces/etc defined in the system.
        `.on(namespace, provider, probe, callback)` with support for wildcards, etc.
  * Use ring buffers and other similar techniques on the consumer side to avoid
  performance hit.
  * Implementation of user side in https://github.com/joyent/node-tracing .
  * Namespace is useful to make sure that there is no conflict between probes,
  and can also allow some optimizations like filtering out what the consumer
  doesn't care about.

### Amendments by Jacob:
  API: do not need to create the probe explicitely. This is done by the API
  the first time a probe is emitted.
  The concern with that (from TJ) is that the consumer of these events often
  needs to know in advance about the probes that can be fired. Not knowing about
  existing probes in advance can make using dtrace and other platform providers
  more difficult (but also other consuming platforms), and as Forest pointed out
  it's a valuable asset for node.

### Concerns
  * This solution is not generic enough to provide a solution for the number one
  priority: Async tracing of any function/callback. Answer to this concern: this
  API can be used as a tool to provide information about more generic stuff like
  handles creations, etc.

  * How to trace what happens in third party modules? Adding static probes to
  third party modules doesn't seem to be realistic for some people (e.g
  concurix).

  * How to attach the metadata that consumers are interested in? How to make
  sure that consumers have enough information? Monkey patching is currently the
  easiest way to do that.

  * Brute force solution for the problem of getting the metadata: emit all the
  data of the context at the firing site with the event. Slightly better
  proposition from Jacob: emit the associated data only if requested by the
  consumer.

  * Idealistic solution: being able to inspect any context within v8 when the
  consumer handles a probe.

Qard:
The challenge that APM vendors faced with coming up with a generic API was that we all had slightly different needs for data collection. That's a big part of why I've been pursuing https://github.com/nodejs/node-eps/pull/56 as a way to apply instrumentation more abstractly.

othiym23:
Thanks for the detailed notes, @misterdjules, and (belated) apologies for tagging you into the discussion!

Trott:
Closing, but by all means re-open or comment if that isn't right.

PR #56, opened 2017-05-06
proposal: Staged require() with lifecycle hooks

Qard:
I'm still working on this, but I think it's time to get some more eyes on it.

This is the EPS continuation of my work in https://github.com/nodejs/node/pull/12349 to attempt to enable an AST transformation pipeline in Node.js core. There's definitely some stuff missing still, and I'll add more over the weekend, but this is where I'm at today. 😸 

@nodejs/diagnostics

### DIFF
--- a/XXX-staged-require-with-lifecycle-hooks.md
+++ b/XXX-staged-require-with-lifecycle-hooks.md
@@ -1,0 +1,192 @@
+| Title  | Staged require() with lifecycle hooks |
+|--------|---------------------------------------|
+| Author | @qard                                 |
+| Status | DRAFT                                 |
+| Date   | 2017-05-05                            |
+
+# Proposal
+
+There are two major purposes to this proposal:
+- To provide module loading lifecycle hooks
+- To split resolving, loading and compilation concerns
+
+# Motivations
+
+Currently some very popular userland modules depend heavily on monkey-patching
+to provide their functionality. This can be unsafe, inflexible and damaging to
+performance.
+
+Some examples include:
+
+- Tracing, debugging and monitoring
+- Code coverage
+- Mocking
+- Transpiling
+
+### Safety
+
+Monkey-patching can be unsafe for many reasons. Sometimes code the behaviour
+of code changes based on parameter length of input functions--express is a good
+example of that. Sometimes patches can interfere with getter/setter behaviour.
+It can be difficult to differentiate between async callbacks that need code to
+maintain barrier linkage and sync callbacks which do not. Some interfaces
+accept multiple function arguments and dynamically identifying which is a
+callback that needs to be instrumented can be unreliable. Any form of queueing,
+even simply putting a callback in an array to be called elsewhere, will break
+continuation linking and requires manual intervention to maintain that linkage
+which is not always possible for dynamic instrumentation.
+
+Most APM providers try/catch constantly because the existing solutions for
+tracking continuation lose context regularly. We've done the best we can with
+monkey-patching, but we've hit a wall and have been pushing for things like
+`async-hooks` to improve the situation. However, `async-hooks` only solves part
+of the problem--linking the callsite to the callback--but context on what was
+happening still requires monkey-patching, and there are plenty of situations
+where context-loss is not the hard part.
+
+### Performance
+
+Monkey-patching is prone to performance issues. The vast majority requires
+wrapping things in closures. More complex instrumentation can often trigger
+deoptimization. Sometimes heavy runtime checking is required with constant
+dynamic function generation and reassignment, which can put a lot of undue
+stress on the optimizer and garbage collector.
+
+If, for example, one could intercept the module loader and apply an AST
+transform before compilation, there's a great deal that could be gained.
+Most Run-time checking would simply disappear. Most closures would disappear.
+Context loss would disappear because the transformer could take advantage of
+lexical scope to describe linkage directly at callsites.
+
+### Flexibility
+
+Many things are just not possible to any reasonable capacity relying purely on
+monkey-patching. Often things which could provide valuable coverage or tracing
+insight are simply not exposed.
+
+An example of a difficult thing to apply tracing hooks to is the timing from
+when a socket first enters JS in `net.js` to when it switches to a TLS socket.
+One could listen for the `secureConnection` event, but there are several
+failure paths between the `TLSSocket` constructor and the event handler being
+triggered, which could result in context loss. One might also think to patch
+`TLSSocket.prototype._init`, however that will be called by both servers and
+clients, which are difficult to differentiate from in this case. To correctly
+identify and correlate a net socket to a corresponding TLS upgrade takes
+extensive monkey-patching with a lot of run-time checks in a major hot-path.
+With an AST transform, a single line of code could be inserted
+[before the `TLSSocket` construction][1] in the `Server` constructor.
+
+### ES Modules
+
+A big issue on the horizon is that, if ES Modules are adopted, modules become
+immutable at build time. This means that the current monkey-patching approach
+will not work anymore. TC39 has already suggested using AST transformation
+to deal with this issue.
+
+Another concern with ES Modules is the difference in loader behaviour. This is
+where splitting resolution and loading concerns from compilation is valuable.
+
+# Implementation
+
+The idea is to go for something similar to the existing `Module._extensions`
+interface, but with three stages: resolving, loading and compiling.
+
+Each stage would interact with a `ModuleRequest` object which expresses the
+lifetime of a module request through the `resolve`, `load` and `compile` stages.
+Initially, it would only include the `parent` and `inputPath` fields. Other
+fields would be populated in the different lifecycle stages.
+
+### `ModuleRequest`
+
+##### `moduleRequest.parent`
+
+The parent module which initiated the `require()` request.
+
+##### `moduleRequest.inputPath`
+
+The input string to the `require()` call or `import` statement which initiated
+this module request.
+
+##### `moduleRequest.resolvedPath`
+
+The fully resolved file path, as determined by the `require.resolvers` handler.
+This would be added to the module request object during the `resolver` stage.
+
+##### `moduleRequest.contents`
+
+The content string (or buffer?) of the module. This field is populated during
+the `loader` stage.
+
+##### `moduleRequest.module`
+
+The final module object, built by the `compiler` stage.
+
+### `require.resolvers`
+
+It is the responsibility of `require.resolvers` extension handlers to receive
+the in-process `Module` object and use the metadata on it to figure out how to
+resolve a load. For `require()` relative location loading, it'd simply look at
+the `parent` and resolve the input string of the `require()` call relative to
+the `parent` module.
+
+This would roughly correspond to the current `Module._resolveLookupPaths`
+implementation, but it should instead receive the module request object which
+it will modify in-place.
+
+```js
+const mockedPaths = {
+  'foo.js': path.join(__dirname, 'mocked-foo.js')
+}
+
+const originalResolver = require.resolvers['require']
+require.resolvers['require'] = function customResolver(request) {
+  if (Object.keys(mockedPaths).includes(request.inputPath)) {
+    return mockedPaths[request.inputPath]
+  }
+  originalResolver(request)
+}
+```
+
+By defining named behaviour specific to `require()`, we open the door for
+supporting the different behaviour of `import` in the future.
+
+### `require.loaders`
+
+The `require.loaders` extension handlers receive the `Module` object after the
+`resolver` stage and use the resolved path to attempt to load the contents of
+the module file.
+
+```js
+const originalLoader = require.loaders['.js']
+require.loaders['.js'] = function customLoader(request) {
+  // This populates the `contents` field
+  originalLoader(request)
+
+  request.contents = request.contents.replace(
+    /module\.exports\s+\=/,
+    // Don't forget that `exports` object now!
+    'exports = module.exports ='
+  )
+}
+```
+
+### `require.compilers`
+
+The `require.compilers` extension handlers receive the `Module` object after the
+`loader` stage and pass the `contents` into a compiler function to build the
+final module object.
+
+```js
+require.compilers['.yaml'] = function yamlCompiler(request) {
+  request.module.exports = yaml.parse(request.contents)
+}
+```
+
+# Notes
+
+- The `ModuleRequest` object could possibly also just be a `Module` instance,
+but the current resolution stage occurs before attempting to build a module
+instance, so it's possible backwards-incompatible changes may be required to
+do that, thus the safer approach of an entirely new object.
+
+[1]: https://github.com/nodejs/node/blob/master/lib/_tls_wrap.js#L829
--- a/XXX-staged-require-with-lifecycle-hooks.md
+++ b/XXX-staged-require-with-lifecycle-hooks.md
@@ -110,13 +110,10 @@ where splitting resolution and loading concerns from compilation is valuable.
 
 # Implementation
 
-The idea is to go for something similar to the existing `Module._extensions`
-interface, but with three stages: resolving, loading and compiling.
-
-Each stage would interact with a `ModuleRequest` object which expresses the
-lifetime of a module request through the `resolve`, `load` and `compile` stages.
-Initially, it would only include the `parent` and `inputPath` fields. Other
-fields would be populated in the different lifecycle stages.
+There are a few possible implementations to discuss, both of which rely on a
+`ModuleRequest` object to express the module loading lifecycle. Initially it
+would only include the `parent` and `inputPath` fields. Other fields would be
+populated in the different lifecycle stages.
 
 ### `ModuleRequest`
 
@@ -143,6 +140,11 @@ the `loader` stage.
 
 The final module object, built by the `compiler` stage.
 
+## Hook-based
+
+The idea here is to go for something similar to `Module._extensions`, but with
+three stages: resolving, loading and compiling.
+
 ### `require.resolvers`
 
 It is the responsibility of `require.resolvers` extension handlers to receive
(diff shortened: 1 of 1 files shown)


XXX-staged-require-with-lifecycle-hooks.md
```diff
@@ -0,0 +1,192 @@
...
+
+- Tracing, debugging and monitoring
+- Code coverage
+- Mocking
+- Transpiling
+
+### Safety
+
+Monkey-patching can be unsafe for many reasons. Sometimes code the behaviour
+of code changes based on parameter length of input functions--express is a good
```
sam-github: I know what you are trying  to say about n-arity of functions, but the grammar is a bit garbled here.

XXX-staged-require-with-lifecycle-hooks.md
```diff
@@ -0,0 +1,192 @@
...
+identify and correlate a net socket to a corresponding TLS upgrade takes
+extensive monkey-patching with a lot of run-time checks in a major hot-path.
+With an AST transform, a single line of code could be inserted
+[before the `TLSSocket` construction][1] in the `Server` constructor.
+
+### ES Modules
+
+A big issue on the horizon is that, if ES Modules are adopted, modules become
+immutable at build time. This means that the current monkey-patching approach
+will not work anymore. TC39 has already suggested using AST transformation
```
sam-github: do you have a link to this discussion, or other ref?

XXX-staged-require-with-lifecycle-hooks.md
```diff
@@ -0,0 +1,192 @@
...
+The content string (or buffer?) of the module. This field is populated during
+the `loader` stage.
+
+##### `moduleRequest.module`
+
+The final module object, built by the `compiler` stage.
+
+### `require.resolvers`
+
+It is the responsibility of `require.resolvers` extension handlers to receive
```
sam-github: Reminds me of https://www.lua.org/manual/5.1/manual.html#pdf-package.loaders

XXX-staged-require-with-lifecycle-hooks.md
```diff
@@ -0,0 +1,192 @@
...
+require.resolvers['require'] = function customResolver(request) {
+  if (Object.keys(mockedPaths).includes(request.inputPath)) {
+    return mockedPaths[request.inputPath]
+  }
+  originalResolver(request)
+}
+```
+
+By defining named behaviour specific to `require()`, we open the door for
+supporting the different behaviour of `import` in the future.
```
sam-github: maybe should be an array of resolvers, where if a resolver fails to resolve, the next one is tried?

sam-github: also, the API only makes sense if in the future there will be a key that is not "require", which is a bit YAGNI, unless you actually demonstrate that the API will work for a future use-case ("import" is what I assume you are leaving the door open for)

XXX-staged-require-with-lifecycle-hooks.md
```diff
@@ -0,0 +1,192 @@
...
+  // This populates the `contents` field
+  originalLoader(request)
+
+  request.contents = request.contents.replace(
+    /module\.exports\s+\=/,
+    // Don't forget that `exports` object now!
+    'exports = module.exports ='
+  )
+}
+```
```
sam-github: will c++ addons go through the loader?

sam-github:
I wonder if a simpler approach of only a single hook would be better, but with Module offering APIs that can be used to implement that hook - APIs like _resolveLookupPaths() (but doced).

The 3-part API reminds me a bit of the mid-layer anti-pattern: https://lwn.net/Articles/336262/

Its very modelled on current behaviour, but if I wanted to implement a resolve like that used in https://github.com/zeit/pkg, I might want to completely replace all 3 stages with a different implementation, and would want just one hook, and then have a library of useful methods (resolve, load, parse, rewrite, ...) that I can use if I want, but no have the assumption that those 3 stages always occur baked into the extension hook API.

^--- not a rejection, just raising the possibility of a different approach that might be worth considering

cc: @watson @nodejs/diagnostics

XXX-staged-require-with-lifecycle-hooks.md
```diff
@@ -0,0 +1,192 @@
...
+
+The `require.compilers` extension handlers receive the `Module` object after the
+`loader` stage and pass the `contents` into a compiler function to build the
+final module object.
+
+```js
+require.compilers['.yaml'] = function yamlCompiler(request) {
+  request.module.exports = yaml.parse(request.contents)
+}
+```
```
sam-github: also seems like it should be a stack. If the code was .ts, it might need compilation to .js, but after it's compiled to js, an APM that instruments js would want to get it so it can be rewritten for instrumentation

Qard:
Yep, I had similar thinking. Hadn't thought yet about how to encode it into the proposal, but basically my thinking was just that the three-stage thing would be what the internals does, but higher levels could be wrapped too, if that's more suitable.

XXX-staged-require-with-lifecycle-hooks.md
Qard: I'll work on it. Thanks for the input. 😅

XXX-staged-require-with-lifecycle-hooks.md
Qard: It came up in one-on-one discussion with @bmeck at one point. Not sure if that is doc'd somewhere.

sam-github:
It might be clearer to remove refs to `AST`, because no AST is exposed anywhere here. APMs would have to use esprima (or just regex or text substitution) to directly rewrite the js source, whether they do that by transforming to AST, modifying, and reencoding as js is an implementation detail if I understand the proposal.

I am generally warming to the idea of exposing hooks into the module require/import system, but also worried a bit about tying the hooks too closely to what we do today, which may be counter productive if the intention is to enable extensibility into future innovative use cases.

XXX-staged-require-with-lifecycle-hooks.md
Qard: Good idea on the failover design. 👍 

Import is indeed what I'm thinking of with the named list design. That one is much more of a "maybe" to me than the other bits right now. I feel like there's something in here that ES Modules can benefit from in regards to the loader, but I'm not quite sure what yet.

sam-github:
@igorklopov would a proposal such as this help with `pkg` in implementing a loader from the deps compiled into the executable?

XXX-staged-require-with-lifecycle-hooks.md
Qard: Not sure. My thinking was the [single line extensions handler](https://github.com/nodejs/node/blob/master/lib/module.js#L598) currently used would become a compiler and it'd skip the loader step. Not sure if that makes sense though.

XXX-staged-require-with-lifecycle-hooks.md
sam-github: I'd go for simpler and more generic unless we are certain this will be helpful for imports, and actually, as much as es mods are a WIP, its probably not worth doing this unless we are _certain_ it works for them, or its guaranteed to be "legacy" in a couple years.

And isn't it 2am in Vancouver? ;-)

XXX-staged-require-with-lifecycle-hooks.md
Qard: Yeah, my original PR went with a stacked design. I've been going back and forth and if that's much better than just wrapping the existing function repeatedly. It's a trade off of making all requires a bit slower with stack checks/iteration or just making patched environments slightly slower.

sam-github: might be possible to optimize inside module. If the array has only one member, the function used to iterate the array can be replaced directly with the only function in the array. For arrays longer than one, we might even be able to pre-compose a function out of the members of the array (basically, cacheing them to avoid repeated iteration).

XXX-staged-require-with-lifecycle-hooks.md
Qard: Almost 3AM. 😅 

Also, my thinking was to put the import stuff in there just to spark some discussion around that. I'm definitely not tied to that part being in the final proposal. I'd really love to hear from other people more deeply involved in the ES Modules discussion than I, since both things are deeply tied to the module loading systems and I'd really like to get some coordination going on here.

XXX-staged-require-with-lifecycle-hooks.md
Qard: I definitely feel like the stacked interface is friendlier. Maybe I'll put that in tomorrow and have a couple alternative suggestions to discuss and narrow down on.

Qard:
Ok, I rewrote most of the doc and added an alternative implementation. (I kind of prefer the alternative, to be honest)

sam-github:
Still looks to me like the mid-layer anti-pattern: it assumes a 4-stage life-cycle, so makes this inflexible when it could be flexible.

What about something more like:
```
const m = require('module');


m.register(tsRequire);

function tsRequire(name, parent) {
  // name is the module name passed to require, parent is the Module doing the require
  // return value is a Module

  if (!name.test(/\.ts$/)) return;
  const found = m.resolveRelative(name, parent); // could look in a zip file or on a web server, or ...
  const src = fs.readFileSync(found);
  const xfrm = ts.compile(src);
  return  m.whatever(xfrm, parent) // however given js source one creates a Module to return
}
```

So there is only one hook, its given what require always knows (name and parent), there are some library functions it can use to do its job (essentially the internal functions that are used by require now), but it can do something _completely_ different if it wants to, as long as the end result is "here is a Module for that name".

Qard:
The problem with that approach is that you can't reasonably intercept the data between the stages. If, for example, I wanted to apply another transform to that code after the ts transpile, I'd have to monkey-patch the ts module itself along with the register function to ensure it only applies my extra transform when using ts through register handlers.

XXX-staged-require-with-lifecycle-hooks.md
bmeck: relevant links:
https://tc39.github.io/ecma262/#sec-module-namespace-exotic-objects
https://tc39.github.io/ecma262/#sec-createimportbinding

sam-github:
@Qard Right, I see your point, and take back my suggestion.

Qard:
I definitely get where you're coming from on wanting a more composable interface though.

Where the proposal is at now tries to balance enough power to cover the particular set of needs we know we have now without sacrificing too much in complexity and potentially performance burden.

I'm definitely open to suggestions on how to improve the proposal. It *is* an EPS after all.

sam-github:
At this point, I think we need more feedback from prospective users.

mhdawson:
Nice write-up with a good overview of the rational.  @sam-github do you have suggestions on who specifically we should ask for feedback ?

bmeck:
If I get time, I would like to review it. That is unlikely prior to June though :(

sam-github:
@mhdawson people who are working in the target use-cases, to quote the proposal:

- Tracing, debugging and monitoring
- Code coverage
- Mocking
- Transpiling

So, APM implementors (new relic, appdynamics, opsbeat, appmetrics, etc), someone from Istanbul (are there other coverage implementations?), mocking... no idea, babel/typescript/...

novemberborn:
[AVA](https://github.com/avajs/ava/) and (nominally) [nyc/Istanbul](https://github.com/istanbuljs/nyc) maintainer here. Thank you @Qard for raising this proposal, we've ended up using packages like [`append-transform`](https://github.com/istanbuljs/append-transform) and [`caching-transform`](https://github.com/avajs/caching-transform) which required a lot of effort to make work correctly. It'd be great to have official support from Node.js itself, especially as ES modules come into play.

`nyc` has a use case where it needs to apply its instrumentation transform last. This is tricky though, since `nyc` bootstraps itself in child process so that it's loaded first, and it is not typically aware of say a user-supplied `babel-register` hook. [`append-transform` goes to great lengths](https://github.com/istanbuljs/append-transform/blob/master/index.js) to achieve this.

Loader objects should make this easier, since it means there's only one list who's order has to be controlled. Ideally the `require.loaders` property would not be configurable or writable. Perhaps it could be a custom collection that controls where in the pipeline a loader is inserted and that prevents arbitrary mutation.

hybrist:
One note about the API: In JS module land, it will also be valuable to control linking. That's the mechanism that would still allow "monkey-patching" (since source transforms have quite a few downsides re: access to dependencies).

bmeck:
@jkrems you need to be careful here, as interrupting linking generally removes exports from being live. There is no notification when exports update and no getter like mechanism upon accessing an import.

hybrist:
@bmeck Interesting! The whole "when is a binding live live" thing still trips me up. What I was thinking was something along the lines of:

```js
// original.js
export function f() {}
export function g() {}
```

```js
// monkey-patch.js
export { g } from '::magic'; // or whatever that syntax is
import { f as originalF } from '::magic';
export function f() {
   // Do typical monkey stuff
   return originalF.apply(this, arguments);
}
```

I definitely have to do more experimentation/reading and it fails terribly if you don't have a full list of the original module's exports (and there currently is no API exposing that info).

bmeck:
@jkrems also, your example is using a fn, which is safe-ish since it evaluates on each invocation, but things like:

```js
export let now = Date.now();
setInterval(() => now = Date.now());
```

where you can't execute code upon access are the real problems.

Note: v8 has rejected implementing getters for variable access due to a number of reasons.

XXX-staged-require-with-lifecycle-hooks.md
```diff
@@ -0,0 +1,330 @@
...
+# Notes
+
+- The `ModuleRequest` object could possibly also just be a `Module` instance,
+but the current resolution stage occurs before attempting to build a module
+instance, so it's possible backwards-incompatible changes may be required to
+do that, thus the safer approach of an entirely new object.
+- For the object-based style, it might also be a good idea to, rather than have
+the `transform` method, have `preCompile` and `postCompile` transformers.
+
+[1]: https://github.com/nodejs/node/blob/master/lib/_tls_wrap.js#L829
```
watson: You probably wanna future proof this link by tying it to a specific git ref (it's already broken) - tip: press `y` while on the github page and it will update the URL for you with a ref

XXX-staged-require-with-lifecycle-hooks.md
```diff
@@ -0,0 +1,330 @@
...
+
+## Hook-based
+
+The idea here is to go for something similar to `Module._extensions`, but with
+three stages: resolving, loading and compiling.
+
+### `require.resolvers`
+
+It is the responsibility of `require.resolvers` extension handlers to receive
+the in-process `Module` object and use the metadata on it to figure out how to
```
watson: When referencing the `Module` object, did you mean `ModuleRequest`?

XXX-staged-require-with-lifecycle-hooks.md
```diff
@@ -0,0 +1,330 @@
...
+and not a `loader`.
+
+### `require.transformers`
+
+The `require.transformers` extension handlers receive the `Module` object after
+the `loader` stage and alter the text content of the module.
+
+```js
+require.transformers.push(({ contents, resolvedPath }) => {
+  if (!/\.js$/.test(resolvedPath)) return
```
watson, line 210: shouldn't it `return contents`? Or is returning `undefined` the same as not modifying the contents?

XXX-staged-require-with-lifecycle-hooks.md
Qard: Yep. I'll make some revisions soon. Pretty busy for the next few days! 😱

XXX-staged-require-with-lifecycle-hooks.md
Qard: Yep, returning undefined means no changes. Could work either way though.

XXX-staged-require-with-lifecycle-hooks.md
Qard: Good catch, I'll fix that soon.

matthewloring:
How will these lifecycle events work with the module cache? Would you expect to do the module cache lookup after resolution as is done currently and if so would only the resolution event be emitted on cache hits?

watson:
@matthewloring Personally I'd expect the cache to be populated with the output of the pipeline. So the cache would then continue to function just as today. I can't currently see a use-case where you'd like any of the hooks to fire if the module is being loaded from the cache... but I might have over looked something

matthewloring:
@watson The cache is currently keyed on the result of the resolution step so I think the user provided resolver would still be run before the cache lookup could occur. The use case that comes to mind for this would be module wrapping where you want the wrapper module to be given when the parent is not the wrapper module itself (which needs a handle to the module being wrapped). The cache does not currently take the parent into account to my knowledge which could interfere with this behavior.

Qard:
I made a few revisions. The `resolve` hook should always run, even when there's a cache entry, since the cache entries are currently structured as resolved paths for keys.

@matthewloring That should be possible by looking at the `parent` property on the `ModuleRequest`.

matthewloring:
Ok, that sound good. It does exclude applying different source transformations based on the parent module but I'm not sure how compelling that use case is.

mcollina:
I do not think we should expose this API through the standard `require` function. Let's attach it somewhere more "hidden", as it is a relative advanced topic, if this lands.

I think we should state that this new API will not have any slowdown in load time if it is not enabled. Node.js fast boot time is one of its main features, and we should keep that. Moreover, it should be fast even it is enabled, as there should be little overhead in passing the data.

Given that ES6 modules are loaded at a different time compared to the standard `require`. ES6 modules are loaded during parsing, rather than at runtime. Could this transformation pipeline work with ES6 modules? In that case, would we be able to implement something like http://npm.im/proxyquire for ES6 modules?

BridgeAR:
There was no progress for a long while here and I do not see any conclusion. Is there any progress here?

Qard:
I'm planning on revisiting this soon and making some updates related to supporting es modules. :)

Trott:
Closing, but feel free to re-open or move to a more appropriate/higher-visibility repository.

PR #58, opened 2017-05-22
Fix header markdown in 002-es6-modules.md

styfle:
GitHub changed their markdown syntax to require a space after the hash marks.


styfle:
Looks like this was fixed already

PR #61, opened 2017-08-09
clarify the exact MIME for .mjs

bmeck:


bmeck:
requested by https://github.com/jshttp/mime-db/pull/88

Trott (approved):
LGTM

Trott:
Arguably out of scope for this document, but if people are looking to us for the answer on MIME type, this is certainly the answer I'd expect us to give. So 👍

ChALkeR:
The first part looks good:
> The MIME used to identify `.mjs` files should be a [web compatible JavaScript MIME Type](https://html.spec.whatwg.org/multipage/scripting.html#javascript-mime-type)

But I'm not sure about this
> preferably `application/javascript`

because I don't know what's going on the `application/javascript` vs `text/javascript` question and which one is preferred now and by whom.

Perhaps /cc @domenic for that.

Note that I'm not implying that I prefer to see `text/javascript` there — I would have asked the exact same question if this change mentioned `text/javascript` as the preferrable mime type.

bmeck:
@ChALkeR we could remove the second sentence?

ChALkeR:
@bmeck I guess we could, this change would make sense even without mentioning the preferred one of those.

That addition of the preferred mime has its value, so if the decision could be made in a reasonable amount of time to favor one of those for some valid reasons, so I would prefer it to be mentioned, but if that won't be achieved — merging this without an exact preferred mimetype LGTM.

ChALkeR (approved):
LGTM, but either after elaborating (not in the patch itself, but in GitHub comments and/or commit comment) why should the mentioned preferred exact type (whatever it would be) be preferred, or with removing the «preferred» part if there would be no informed decision on that in a reasonable time.

domenic:
text/javascript is preferred on the web generally; I believe some RFC tried to obsolete it, but it's the most prevalent on the web, so the HTML spec just made it the default.

bmeck:
linked to HTML spec and updated

thefourtheye (approved):

ChALkeR:
@bmeck I am still not entirely sure if the HTML spec is the correct source for this.
What do other members of TC39 think? Is [this opinion](https://mail.mozilla.org/pipermail/es-discuss/2011-August/016267.html) by @BrendanEich actual?

ChALkeR:
* ECMAScript spec doesn't specify a mime type.
* IANA [specifies](https://www.iana.org/assignments/media-types/application/javascript) `application/javascript`, and [calls](https://www.iana.org/assignments/media-types/text/javascript) `text/javascript` «obsolete», but IANA isn't a very actual/up-to-date source.
* HTML spec specifies `text/javascript` ~~and calls `application/javascript` «legacy»~~ (_**upd:** not any more_), but HTML spec probably _shouldn't_ be the primary source for JavaScript mime type.
* [RFC 4329](https://tools.ietf.org/html/rfc4329) specifies `application/javascript` as the primary and calls `text/javascript` «obsolete», but that's not even a spec (and it's 11 years old).
* @BrendanEich from TC39 [wrote](https://mail.mozilla.org/pipermail/es-discuss/2011-August/016267.html) 6 years ago that the correct one is `application/javascript` (referring to RFC 4329) and that `text/` isn't even the correct category and should not be used.
* @domenic from TC39 says that `text/javascript` is correct one today and that `application/javascript` is not, and refers to the HTML spec.

domenic:
> HTML spec probably shouldn't be the primary source for JavaScript mime type.

I dunno. HTML is the spec for web browsers, and web browsers are the major client that (a) cares about MIME types and (b) executes JavaScript. So I think it's a pretty good place for it.

refack (approved):

bmeck:
I would agree with @domenic here

bmeck:
@ChALkeR are you ok with `text/javascript` since it is a preference, and it is not a mandatory thing? Is it ok to merge this PR?

gibfahn (approved):

TimothyGu (approved):

bmeck:
landed in https://github.com/nodejs/node-eps/commit/6eef91d4ab4544ee5ed01df86dd7d055e90e130e

ChALkeR:
@bmeck Sorry for a very delayed response on this :disappointed:, I had some personal availability issues, hopefully resolved now.

The PR as landed in 6eef91d looks good to me.

PR #62, opened 2017-08-09
007: --entry-extension CLI flag proposal

bmeck:


bmeck:
@rauschma: [bikeshed for `--input-extension` instead](https://twitter.com/rauschma/status/895303986240253953)

ljharb (changes requested):
What happens with `node --entry-extension=js app.js` or `node --entry-extension=mjs app.js`?

I would expect either an exception when there's a mismatch, or, that the entry-extension would silently override the file extension.

bmeck:
Mixing anything that is not an opaque text stream such as a filename should error since it doesn't make sense.

matthewp:
In @ljharb's case I would want/expect the entry-extension to silently override the file extension. If it throws, then

```shell
node --entry-extension=mjs app.js
```

and 

```shell
node --entry-extension=mjs < app.js
```

Would do different things, and why should they?

matthewp:
Ah, I missed @bmeck's reasoning, nevermind. That's fine I think.

rauschma:
@bmeck: I mostly agree, but executing a `.js` file as an ES module may be a use case.

bmeck:
@rauschma pipe it in over STDIN

rauschma:
@bmeck As that is the only use case I can think of, that’s probably the right call, yes.

ljharb:
As long as it's a clear decision one way or the other, I'm happy :-) it just can't be unspecified.

007-extension-flag.md
```diff
@@ -0,0 +1,43 @@
...
+node --entry-extension=mjs -e $(cat <app.mjs)
+```
+
+## Considerations
+
+If mixing `--entry-extension` with a non-stream argument such as a file path or entry URL, Node should exit with an error prior to any evaluation. This also means that `--entry-extension` cannot be used to override the extension of file paths.
+
+```sh
+# exits with an error
+node --entry-extension=json app.js
```
ljharb, line 30: `node --entry-extension=js app.js` should then also error?

bmeck: yes

ljharb: Might be a good example to add so that people don't think it's the mismatch that errors

matthewp: Definitely a good example since most of the feedback so far has been above this :)

bmeck: added

ljharb (commented):
What happens when a leading `.` is supplied?

What happens when an unknown extension is supplied?

Is there a way to list supported extensions on the CLI?

bmeck:
> What happens when a leading . is supplied?

Stream goes through CJS `require.extensions` then ESM import mechanics. It wouldn't find a `..js` etc. extension and would fail.

> What happens when an unknown extension is supplied?

Fails to load.

> Is there a way to list supported extensions on the CLI?

Not currently, but could be proposed.

bmeck:
This change does not attempt to change workflow of how files are loaded. It just declares the mode rather than going through CJS always. See https://github.com/nodejs/node/blob/5fd2f03b16bfd6730fe82dc3217f1d05131b9615/lib/internal/bootstrap_node.js#L435 for some existing implementation details.

ljharb:
It seems like `node --entry-extension=foo` should fail immediately if `foo` isn't in `require.extensions`, and ideally would print out `Object.keys(require.extensions)` in the error message.

bmeck:
@ljharb I am hesitant to do that error behavior since people could certainly add to the list using `--require`. Also, `require.extensions` is not used in the ESM logic.

ljharb:
@bmeck in that case i'd expect the list to be runtime-computed, so if they'd added to the list, it would become a valid option.

However it's implemented, I think it'll be important to fail-fast when an invalid extension is supplied, and make sure the error message is meaningful and actionable.

007-extension-flag.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Userland Extensions
+
+Some libraries extend `require.extensions` with new extensions. These libraries should be considered when determining the list of extensions allowed in `--entry-extension`.
+
+Any future extensions to the ESM based loading mechanisms will have to account for collisions and define behavior when they collide.
+
+### Early Errors
+
+Any unknown file extension should prevent the main entry point to the program from executing anything. Preloaded files such as using `--require` should continue to load normally.
```
ljharb, line 43: Is there any concept of preloading files, but with a different entry extension? In other words, `--import` to match `--require`?

ljharb (approved):

007-extension-flag.md
bmeck: @ljharb not currently, use a wrapper for now. `--require` uses file paths always, so file extensions are always included. There are some odd problems with naming here, and some timing differences due to ESM being async that I think are fine to put off for a future time.

ljharb: ah, so `--require file.mjs` would work as expected?

bmeck: maybe? `require('./foo.mjs')` throws so the naming is a bit odd but since CLI arg for main file to run accepts `.mjs` it should work fine I would expect.

007-extension-flag.md
```diff
@@ -0,0 +1,57 @@
+| Title  | Implement --extension flag  |
+|--------|-----------------------------|
+| Author | @bmeck                      |
+| Status | DRAFT                       |
+| Date   | 2017-08-09T10:00:00-05:00   |
+
+## Description
+
+The Node CLI currently accepts input in both filename (`node app.js`) and opaque text stream form (`node -e '123'` and `node <app.js`). This behavior always results in running the stream as a CommonJS goal. With the upcoming usage of ESM and WASM there will be cases where developers wish to pipe other forms of text to the CLI.
```
Fishrock123, line 9: `Script goal` and maybe also `other program forms to the CLI`

007-extension-flag.md
```diff
@@ -0,0 +1,57 @@
...
+
+This proposal is to implement an `--entry-extension` CLI flag for the cases which may want to use a stream that does not have a file extension.
```
Fishrock123, line 11: If we are defining the name here, might be worthwhile to also specify `--ext` as a shorter-hand. People will want to see using it as reasonable UX.

007-extension-flag.md
```diff
@@ -0,0 +1,57 @@
...
+```sh
+# exits with an error
+node --entry-extension=json app.js
+# exits with an error
+node --entry-extension=js app.js
+```
+
+### Userland Extensions
+
+Some libraries extend `require.extensions` with new extensions. These libraries should be considered when determining the list of extensions allowed in `--entry-extension`.
```
Fishrock123, line 37: How would this work? Only for preload modules I suppose?

007-extension-flag.md
Fishrock123: Correct, but `require()` has synchronous code boundaries to deal with. We have more freedom in how we define preload's interactions.

Perhaps we should have `--import` alias to `--require` though, for posterity and to avoid confusion. That being said, `-i` is already used.

007-extension-flag.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Early Errors
+
+Any unknown file extension should prevent the main entry point to the program from executing anything. Preloaded files such as using `--require` should continue to load normally.
+
+Any error produced by this should describe how to get the list of known extensions; and the error should attempt to point towards how to add an extension if needed. The error may change from preferring `require.extensions` to a new mechanism that supports ESM over time.
+
+## Alternate Possibilities
+
+Some alternate possibilities exist that might be relevant. `--entry-url` for example could be used to provide an `import.meta.url` properly while also providing the pathname that contains an extension. However, since URLs are not mandated to have file extensions this might be for nothing. Applications can also access `process.cwd()` to recreate similar data to `import.meta.url`.
```
Fishrock123, line 49: Seems far less intuitive?

Fishrock123 (commented):
Seems good overall. Have some comments.

I think this _could_ be a place to define how `-r` works with non-Script goal programs, but I wouldn't be opposed to having that discussion in a different place either.

007-extension-flag.md
Fishrock123: On second thought, I suppose that last one could also be clarified as `source text` if that is more consistent.

007-extension-flag.md
bmeck: do we have any abbreviates `--` flags?

007-extension-flag.md
bmeck: Yep. Only preload would affect this.

007-extension-flag.md
bmeck: yup, not suggested.

007-extension-flag.md
bmeck: IDK, `--import` seems fine and can use the module resolution of ESM rather than CJS *then* ESM

007-extension-flag.md
Fishrock123: Unsure but I don't really see an issue with it.

targos: What about `-x` ?

Fishrock123: Could use that, I suppose. Does it have any common UNIX meaning it might be confused with?

bmeck: no , http://tldp.org/LDP/abs/html/standard-options.html

MylesBorins:
@bmeck do you think it makes sense to close this as we are no longer using the eps process?

bmeck:
@MylesBorins yes, as I presumed all EPs would be considered archived I didn't touch it.

styfle:
### 007 is dead



See what I did there? 😄 

*......I'll show myself out now* 🏃 🚪

